PR #1, opened 2017-11-06
Initial release

pvdlg:
Implementation as discussed in semantic-release/semantic-release#484.

@gr2m can you enable Codecov and Greenkeeper for the repo and add `GH_TOKEN` and `NPM_TOKEN` to Travis ? Thanks

Fix semantic-release/semantic-release#126


gr2m:
Codecov & Greenkeeper enabled, tokens set

lib/get-client-config.js
```diff
@@ -0,0 +1,26 @@
+module.exports = config => {
+  // Form https://github.com/npm/npm/blob/master/lib/cache/caching-client.js#L194
```
gr2m: maybe add permanent link? https://github.com/npm/npm/blob/d081cc6c8d73f2aa698aab36605377c95e916224/lib/cache/caching-client.js#L194

When you press `y` on any file page, it will redirect to permalink

lib/get-client-config.js
```diff
@@ -0,0 +1,26 @@
...
+    retry: {
+      retries: parseInt(config.get('fetch-retries')),
+      factor: parseInt(config.get('fetch-retry-factor')),
+      minTimeout: parseInt(config.get('fetch-retry-mintimeout')),
+      maxTimeout: parseInt(config.get('fetch-retry-maxtimeout')),
+    },
+    userAgent: config.get('user-agent'),
+    defaultTag: config.get('tag'),
+    couchToken: config.get('_token'),
+    maxSockets: parseInt(config.get('maxsockets')),
```
gr2m: add radix argument to all `parseInt` statements. I’m surprised the linking doesn’t catch that. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Octal_interpretations_with_no_radix

lib/get-last-release.js
```diff
@@ -0,0 +1,40 @@
+const {promisify} = require('util');
+const {resolve} = require('url');
```
gr2m: maybe change this to `{resolve: urlResolve}`. Line 16 is rather confusing with the `promisify` and then the `resolve` call, I was looking for a Promise resolve function :)

lib/get-last-release.js
```diff
@@ -0,0 +1,40 @@
...
+
+module.exports = async ({publishConfig, name}, logger) => {
+  const config = npmConf();
+  const tag = (publishConfig || {}).tag || config.get('tag');
+  const {NPM_TOKEN, NPM_USERNAME, NPM_PASSWORD, NPM_EMAIL} = process.env;
+  const client = new RegClient(getClientConfig(config));
+  const registry = await getRegistry(publishConfig, name);
+
+  try {
+    const data = await promisify(client.get.bind(client))(resolve(registry, name.replace('/', '%2F')), {
```
gr2m: this is quite a lot at once to read easily, can you maybe split it up into multiple lines? At least move the value that `resolve(...)` returns into an own variable for better readability

lib/get-last-release.js
```diff
@@ -0,0 +1,40 @@
...
+      auth: NPM_TOKEN ? {token: NPM_TOKEN} : {username: NPM_USERNAME, password: NPM_PASSWORD, email: NPM_EMAIL},
+    });
+    if (data && !data['dist-tags']) {
```
gr2m: why do you check for `data` her? Should this maybe be `!data || !data['dist-tags']`?

lib/set-npmrc-auth.js
```diff
@@ -0,0 +1,26 @@
...
+const nerfDart = require('nerf-dart');
+const SemanticReleaseError = require('@semantic-release/error');
+const getRegistry = require('./get-registry');
+
+module.exports = async ({publishConfig, name}, logger) => {
+  const registry = await getRegistry(publishConfig, name);
+  logger.log('Verify authentication for registry %s', registry);
+  const {NPM_TOKEN, NPM_USERNAME, NPM_PASSWORD, NPM_EMAIL} = process.env;
+
+  if (!getAuthToken(registry)) {
```
gr2m: maybe return early instead of wrapping the rest of the code into the if statemen?

```js
if (getAuthToken(registry)) {
  return
}
```

lib/set-npmrc-auth.js
```diff
@@ -0,0 +1,26 @@
...
+    if (NPM_USERNAME && NPM_PASSWORD && NPM_EMAIL) {
+      // Using the old auth token format is not considered part of the public API
+      // This might go away anytime (i.e. once we have a better testing strategy)
```
gr2m: can we drop support for `NPM_USERNAME` & `NPM_PASSWORD` and only support `NPM_TOKEN`?

lib/get-client-config.js
pvdlg: Yep.

There is an eslint rule for that: https://eslint.org/docs/rules/radix
But it's not enabled in `standard`: https://github.com/standard/eslint-config-standard/blob/master/eslintrc.json

This is one a reason I prefer other solution to `standard`, they enable most style rules and disable a lot of really useful rules to detect these type of bug...Anyway.

lib/get-last-release.js
pvdlg: This is something that come from the old plugin. It's to handle a weird case that happen on npm when the call return `data` but no `dist-tags`.
Normally when a package doesn't exist it returns a 404.
If I remember well this situation happens when you publish a version with a scope and you unpublish it, the registry return `{}` instead of a 404.

lib/set-npmrc-auth.js
pvdlg: No unfortunately....It's because https://github.com/npm/npm-registry-couchapp doesn't accept login with a token. It accept only the legacy auth (username:password in base 64 + email).

I couldn't find a way to make https://github.com/npm/npm-registry-couchapp authentication works like https://www.npmjs.com/ with a token.

lib/get-client-config.js
pvdlg: Nice, I didn't know that :)

lib/set-npmrc-auth.js
pvdlg: Done

lib/get-last-release.js
pvdlg: Done

lib/get-last-release.js
pvdlg: Done

lib/get-client-config.js
pvdlg: Done

lib/get-client-config.js
pvdlg: Done

Issue #6, opened 2017-11-29
No way to configure that we should always authenticate to NPM 


pbomb:
My scenario is that our NPM server is a locally-running Artifactory instance and every interaction with is requires authentication - even reads. I'm getting a 403 during the call to get the last release. This call configures the npm client with the `auth` property, but does not set `alwaysAuth: true` and does not provide a way for me to configure this plugin to do that. I've tested it locally and adding that option solves my issue.

I assume that we don't want to set always set that to true for all users, but at least providing a way to configure this plugin to operate in that mode would help me. I could start working on a pull request if you think adding this is the right thing to do. Thanks.

pvdlg:
Thanks for the report!
I think you can configure this option in a `.npmrc` file at the root of your project:
```
always-auth = true
```

The config is read [get-last-release.js#L9](https://github.com/semantic-release/npm/blob/master/lib/get-last-release.js#L9).
We prefer to configure all `npm` options in the `.npmrc` rather than adding new options to the plugin.

Let us know if that work. If it doesn't work, the proper fix will be to make sure we read this property from `.npmrc`.

On a side note it would be good to add some comments about that in the README.

pvdlg:
Looking at it more in details the `alwaysAuth` at to be configured in the `auth` object when calling the client and not in the regular config.

So the client call should something along those lines:
```js
    const data = await promisify(client.get.bind(client))(
      uri,
      Object.assign(
        {auth: NPM_TOKEN ? {token: NPM_TOKEN} : {username: NPM_USERNAME, password: NPM_PASSWORD, email: NPM_EMAIL}},
        {alwaysAuth: config.get('always-auth')}
      )
    );
```

PR with tests is welcome!

pbomb:
Thanks for taking a look. I tested your latest code snippet and it does work! I didn't realize I could pull the `always-auth` field from the config. Thanks for the tip. I'll start working on a PR.

pbomb:
FYI, I created https://github.com/semantic-release/npm/pull/7 to address this issue.

pvdlg:
Are you authenticating with a token or with username / password and email? Just wondering.

Looking at the client code it seems that in the case of a call that set the `token` the request is always authenticated no matter the `always-auth` parameter.

[...]

PR #10, opened 2017-11-30
Support legacy authentication

pvdlg:
We are supporting legacy npm authentication as we use [npm-registry-couchapp](https://github.com/npm/npm-registry-couchapp) for integration tests which doesn't support token authentication.
But it wasn't intended to be used in production.

It seems there is several private npm registry solution that supports only legacy authentication.
See semantic-release/semantic-release#500 and #6.

This PR improve our support of the legacy auth and mention it in the documentation.

Fix semantic-release/semantic-release#500

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,26 @@ Publish the package on the `npm` registry.
 
 ## Configuration
 
-For each plugin, the `npm` authentication token has to be configured with the environment variable `NPM_TOKEN`.
+### Environment variables
 
-All the plugins are based on `npm` and will use the configuration from `.npmrc`. Any parameter returned by `npm config list` will be used by each plugin.
+The `npm` authentication configuration is **required** and can be set via environment variables.
 
-The registry and dist-tag can be configured in the `package.json` and will take precedence on the configuration in `.npmrc`:
+Both the [token](https://docs.npmjs.com/getting-started/working_with_tokens) and the legacy (`username`, `password` and `email`) authentication are supported. It is recommended to use the [token](https://docs.npmjs.com/getting-started/working_with_tokens) authentication. The legacy authentication is supported as the alternative npm registries [Artifactory](https://www.jfrog.com/open-source/#os-arti) and [npm-registry-couchapp](https://github.com/npm/npm-registry-couchapp) only supports that form of authentication at this point.
+
+| Variable       | Description                                              
+| -------------- | ----------------------------------------------------------------------------------------------------------------------------- |
+| `NPM_TOKEN`    | Npm token created via [npm token create](https://docs.npmjs.com/getting-started/working_with_tokens#how-to-create-new-tokens) |
+| `NPM_USERNAME` | Npm username created via [npm adduser](https://docs.npmjs.com/cli/adduser) or on [npmjs.com](https://www.npmjs.com)           |
+| `NPM_PASSWORD` | Password of the npm user.                                                                                                     |
+| `NPM_EMAIL`    | Email address associated with the npm user                                                                                    |
+
+Use either `NPM_TOKEN` for token authentication or `NPM_USERNAME`, `NPM_PASSWORD` and `NPM_EMAIL` for legacy authentication
+
+### Options
+
+The plugins are based on `npm` and will use the configuration from [`.npmrc`](https://docs.npmjs.com/files/npmrc). See [npm config](https://docs.npmjs.com/misc/config) for the option list.
+
+The [`registry`](https://docs.npmjs.com/misc/registry) and [`dist-tag`](https://docs.npmjs.com/cli/dist-tag) can be configured in the `package.json` and will take precedence over the configuration in `.npmrc`:
 ```json
 {
   "publishConfig": {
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -20,11 +20,26 @@ Publish the package on the `npm` registry.
...
 ## Configuration
 
-For each plugin, the `npm` authentication token has to be configured with the environment variable `NPM_TOKEN`.
+### Environment variables
 
-All the plugins are based on `npm` and will use the configuration from `.npmrc`. Any parameter returned by `npm config list` will be used by each plugin.
+The `npm` authentication configuration is **required** and can be set via environment variables.
 
-The registry and dist-tag can be configured in the `package.json` and will take precedence on the configuration in `.npmrc`:
+Both the [token](https://docs.npmjs.com/getting-started/working_with_tokens) and the legacy (`username`, `password` and `email`) authentication are supported. It is recommended to use the [token](https://docs.npmjs.com/getting-started/working_with_tokens) authentication. The legacy authentication is supported as some npm private registry do not suport the token authentication.
```
gr2m: maybe instead of 

> some npm private registry

it would be better to list the one we know? I think it’s only Artifactory right now? Maybe sth like this?

> The legacy authentication is supported as the alternative npm registry [Artifactory](https://www.jfrog.com/open-source/#os-arti) only supports that form of authentication at this point.

pvdlg: There is also https://github.com/npm/npm-registry-couchapp.

gr2m: how about tthis

> The legacy authentication is supported as the alternative npm registries [Artifactory](https://www.jfrog.com/open-source/#os-arti) and [npm-registry-couchapp](https://github.com/npm/npm-registry-couchapp) only supports that form of authentication at this point.

pvdlg: I made the change

pvdlg:
I keep getting a lot of `Couldn't start npm-docker-couchdb after 2 min`.
It happened to me 3 or 4 times. Each time on Node 9. While the Node 8 build was passing...

That's going to be a tricky one to figure out...

gr2m:
yeah debugging CI is the worst :(

pvdlg:
I check the availability of the the registry [here](https://github.com/semantic-release/npm/blob/master/test/helpers/npm-registry.js#L34).
Basically I wait for `/registry/_design/app` to return a 200...So not sure what could go wrong...

SimenB:
I'm trying to make this work, perfect timing :D

Have you tested this against artifactory? I can't seem to make it work properly, but I don't get why it's failing. Tried `semantic-release --debug` but that didn't help much. AFAICT I'm on the latest of everything as of 10 minutes ago (including 2.2.0 of this module).

image

Variables part:

```
Setting environment variables from repository settings
$ export NPM_TOKEN=[secure]
$ export ARTIFACTORY_CONTEXT=https://artifacts.schibsted.io/artifactory
$ export ARTIFACTORY_NPM_SECRET=[secure]
$ export [secure]=[secure]
$ export ARTIFACTORY_USER=[secure]
$ export AWS_ACCOUNT_NAME=infra-pro
$ export SPT_ENGPROD_REPORT_KEY=[secure]
$ export SPT_ENGPROD_REPORT_SECRET=[secure]
Setting environment variables from .travis.yml
$ export GITHUB_URL=https://github.schibsted.io
$ export GITHUB_PREFIX=https://github.schibsted.io/api/v3
$ export NPM_EMAIL=$ARTIFACTORY_USER
$ export NPM_OLD_TOKEN=$ARTIFACTORY_NPM_SECRET
$ export GH_TOKEN=[secure]
```

Any ideas about what I should try to debug this?

For all I know the token really isn't valid, I haven't used it before.

But it would be great if `--debug` added some information that it actually tried to use legacy instead of new one, etc.

(Artifactory version 4.16.0)

pvdlg:
The logs mention `Wrote NPM_TOKEN to .npmrc` which mean you have an environment variable called `NPM_TOKEN` and therefore the plugin uses the token authentication.

From the README:

> Use either NPM_TOKEN for token authentication or NPM_USERNAME, NPM_PASSWORD and NPM_EMAIL for legacy authentication

SimenB:
Yeah, I just noticed that... But that is set globally on Travis for us, so I'd be surprised if it was wrong. I can ask the guys who put it there, but is there anything I can do myself to debug more?

SimenB:
https://github.com/semantic-release/npm/blob/5fb0b09f61ad229e5201d803050fd5fbd51e4937/lib/verify.js#L9-L12

For all I know I get a 404 or something, some more information than "no matter what fails with `whoami` it's auth failure".

Or maybe I'm missing some other variable so it doesn't use the correct registry

pvdlg:
It seems you want to configure the legacy auth. But you have an `NPM_TOKEN` which tells the plugin  to use the token auth.

If you want to use the legacy auth configure it according to the doc with `NPM_USERNAME `, `NPM_PASSWORD` and `NPM_EMAIL`, without `NPM_TOKEN`.

SimenB:
A `- export NPM_TOKEN=""` seems to have done it! And not calling it `NPM_OLD_TOKEN`, but `NPM_PASSWORD`. I'm now getting an error from calling github, but that's not related to this repo.

Thanks for the help!

pvdlg:
Probably because of your `GITHUB_PREFIX` configuration. It should be a prefix and not the full URL.

SimenB:
Yup, I noticed, thanks! Opened up https://github.com/semantic-release/github/issues/10 for it.

Now I'm stuck at https://github.com/semantic-release/semantic-release/issues/540, but I feel like I'm getting closer at least

PR #12, opened 2017-12-02
Add `npmPublish` and `tarballDir` options

pvdlg:
Add 2 new options:
- `npmPublish` can be set to `false` to skip the publishing on npm registry
- If `tarballDir` is set the npm tarball (`npm pack`) will be generated in the configured directory

That will allow to users who do not want to publish on the registry to still use the plugin to:
- Update the `package.json` and publish it with [git]([npm](https://github.com/semantic-release/git)) or [github]([npm](https://github.com/semantic-release/github))
- To generate the npm tarball and publish it as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,17 +16,19 @@ Determine the last release of the package on the `npm` registry.
 
 ## publish
 
-Publish the package on the `npm` registry.
+Update the `package.json` version, [create](https://docs.npmjs.com/cli/pack) the `npm` package tarball and [publish](https://docs.npmjs.com/cli/publish) to the `npm` registry.
 
 ## Configuration
 
-### Environment variables
+### Npm registry authentication
 
-The `npm` authentication configuration is **required** and can be set via environment variables.
+The `npm` authentication configuration is **required** and can be set via [environment variables](#environment-variables).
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -63,3 +73,29 @@ Each individual plugin can be disabled, replaced or used with other plugins in t
...
+        "path": "@semantic-release/git",
+        "assets": "dist/*.tgz"
+      },
+      {
+        "path": "@semantic-release/github",
+        "assets": "dist/*.tgz"
+      },
+    ]
+  }
+}
```
gr2m: This will commit the tarbal to git as well as upload it to the  release on GitHub, right? I’m not sure if that makes a lot of sense? Wouldn’t just the upload to the github release be the more common use case?

pvdlg: > This will commit the tarbal to git as well as upload it to the release on GitHub, right?

Yes

> I’m not sure if that makes a lot of sense? Wouldn’t just the upload to the github release be the more common use case?

The intention was to demonstrate the capability more than recommending a specific configuration. But that might be confusing indeed.
I'll change it to upload only to only upload to github.

pvdlg: Done

PR #14, opened 2017-12-02
fix: set LEGACY_TOKEN as auth

### LINKED ISSUE #13: Unable to get auth working against artifactory
I get a 401 when semantic-release tries to get the latest version of the package.

```
[Semantic release]: Call plugin get-last-release
info attempt registry request try #1 at 10:14:18
http request GET https://artifacts.schibsted.io/artifactory/api/npm/npm-virtual/@finn-no%2Ftest-semantic-release
http 401 https://artifacts.schibsted.io/artifactory/api/npm/npm-virtual/@finn-no%2Ftest-semantic-release
[Semantic release]: An error occurred while running semantic-release: { Error: Registry returned 401 for GET on https://artifacts.schibsted.io/artifactory/api/npm/npm-virtual/@finn-no%2Ftest-semantic-release
```

I set up auth like this: `curl -u $ARTIFACTORY_USER:$ARTIFACTORY_PWD https://artifacts.schibsted.io/artifactory/api/npm/npm-virtual/auth/finn-no > .npmrc`. I know this works as `yarn` right after it (which includes fetching dependencies only available on that Artifactory instance) succeeds.

I then do this (horribly inefficient, but it works):
```
- export NPM_TOKEN=""
- export NPM_PASSWORD=`node -p "require('npm-conf')().get('//artifacts.schibsted.io/artifactory/api/npm/npm-virtual/:_password');"`
- export NPM_USERNAME=`node -p "require('npm-conf')().get('//artifacts.schibsted.io/artifactory/api/npm/npm-virtual/:username');"`
- export NPM_EMAIL=`node -p "require('npm-conf')().get('//artifacts.schibsted.io/artifactory/api/npm/npm-virtual/:email');"`
```
in order to make the same auth available for semantic-release. (I have verified by `cat` that these variables are set).

Doing `npm info` against the same package after semantic-release gets 401 results in the expected 404 against the same url. So somehow auth is not provided to the npm plugin properly.

Any ideas on what's going wrong or how I can debug?

See https://github.com/semantic-release/travis-deploy-once/pull/27 for previous discussion

SimenB:
Fixes #13

Finally had a successful publish :D

The real fix might be to pull auth from `.npmrc`, but this works.

These are the overrides in my package.json:

```json
"resolutions": {
  "@semantic-release/condition-travis": "SimenB/condition-travis#db3fd849cdf2ad9162f2acdf5f409f9e82f5ac76",
  "@semantic-release/npm": "SimenB/npm-1#c5e894b513886a4fbceafc3532b55a5ef859f5bc",
  "travis-ci": "pwmckenna/node-travis-ci#39ad0c315e53bc360850994ebaf01f73dc14a18e"
}
```
The 2 travis ones are for support of Travis Enterprise, while the npm one is the commit in this PR.

This is a successful publish to Artifactory 4.16.0 from Travis Enterprise, publishing a release on GitHub Enterprise.

### DIFF
--- a/lib/get-last-release.js
+++ b/lib/get-last-release.js
@@ -8,13 +8,15 @@ const getRegistry = require('./get-registry');
 module.exports = async ({publishConfig, name}, logger) => {
   const config = npmConf();
   const tag = (publishConfig || {}).tag || config.get('tag');
-  const {NPM_TOKEN, NPM_USERNAME, NPM_PASSWORD, NPM_EMAIL} = process.env;
+  const {NPM_TOKEN, NPM_USERNAME, NPM_PASSWORD, NPM_EMAIL, LEGACY_TOKEN} = process.env;
   const client = new RegClient(getClientConfig(config));
   const registry = await getRegistry(publishConfig, name);
 
   try {
     const uri = urlResolve(registry, name.replace('/', '%2F'));
-    const auth = NPM_TOKEN ? {token: NPM_TOKEN} : {username: NPM_USERNAME, password: NPM_PASSWORD, email: NPM_EMAIL};
+    const auth = NPM_TOKEN
+      ? {token: NPM_TOKEN}
+      : {username: NPM_USERNAME, password: NPM_PASSWORD, email: NPM_EMAIL, auth: LEGACY_TOKEN};
     auth.alwaysAuth = config.get('always-auth');
     const data = await promisify(client.get.bind(client))(uri, {auth});
     if (data && !data['dist-tags']) {


lib/get-last-release.js
```diff
@@ -8,13 +8,15 @@ const getRegistry = require('./get-registry');
...
   const client = new RegClient(getClientConfig(config));
   const registry = await getRegistry(publishConfig, name);
 
   try {
     const uri = urlResolve(registry, name.replace('/', '%2F'));
-    const auth = NPM_TOKEN ? {token: NPM_TOKEN} : {username: NPM_USERNAME, password: NPM_PASSWORD, email: NPM_EMAIL};
+    const auth = NPM_TOKEN
+      ? {token: NPM_TOKEN}
+      : {username: NPM_USERNAME, password: NPM_PASSWORD, email: NPM_EMAIL, auth: LEGACY_TOKEN};
     auth.alwaysAuth = config.get('always-auth');
```
SimenB: this is too naive, it's needs to check for `always-auth` on both root and scoped.

See https://github.com/npm/npm/blob/0cc9d89ed2d46745f91d746fda9d205fd39d3daa/lib/config/get-credentials-by-uri.js#L24-L29

For now, I just add it in my .npmrc on root as well

pvdlg:
This is not the right fix. The `auth` property is set [here](https://github.com/npm/npm-registry-client/blob/master/lib/authify.js#L19) by the client in case the authentication is required.

The change in this PR will ignore the `always-auth` parameter and always force the authentication.
That might for your case but likely break for everyone else.

The problem is that `always-auth` is not read properly and `false` is passed to the client.

You have to add to the `.npmrc` file at the root of your repo:
```
always-auth = true
```

lib/get-last-release.js
pvdlg: I tried to figure out how we can end up in a situation with something like that in `.npmrc`:
```
//<registry_url>/:alwaysAuth=true
```
But I couldn't figured out how any `npm` CLI command would result in that.

What's the npm command that write this config?

SimenB: It's returned when authing against artifctory, at least https://www.jfrog.com/confluence/display/RTF/Npm+Registry#NpmRegistry-UsingBasicAuthentication

I couldn't get the npm client to write it, but it does support reading it

SimenB:
It seems you are correct! I probably had some weird auth setup previously and in the process of logging all the things, something.

I can send a PR for the `always-auth` stuff, but publishing works without this PR.

PR #15, opened 2017-12-03
feat: get always-auth from scoped

SimenB:
As mentioned in #13, the npm client supports scoped configuration in npmrc. This PR ports that functionality over.

I opened up https://github.com/kevva/npm-conf/pull/10, but I'm not sure if it will be accepted

### DIFF
--- a/lib/get-last-release.js
+++ b/lib/get-last-release.js
@@ -4,6 +4,18 @@ const npmConf = require('npm-conf');
 const RegClient = require('npm-registry-client');
 const getClientConfig = require('./get-client-config');
 const getRegistry = require('./get-registry');
+const toNerfDart = require('./nerf-dart');
+
+function getAlwaysAuth(registry, config) {
+  const nerfed = toNerfDart(registry);
+  const registryAuth = config.get(`${nerfed}:always-auth`);
+
+  if (registryAuth !== undefined) {
+    return registryAuth;
+  }
+
+  return config.get('always-auth');
+}
 
 module.exports = async ({publishConfig, name}, logger) => {
   const config = npmConf();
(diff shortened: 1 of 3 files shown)


pvdlg:
I don't think this is something we should add. It's a very very edge case, that's probably related to legacy code in `npm`. I couldn't find any way to make the `always-auth` param registry scoped via `npm config set`.

If we were to support it I think it should be done in kevva/npm-conf, so when we call `config.get('always-auth')` it would look for both the scoped and non scoped value based on the registry configured.

SimenB:
All configuration in npm can be scoped to certain registries, or be "global". Nothing edge case about that.

Both yarn and npm respect it (without making the config loader care about it) (https://github.com/yarnpkg/yarn/blob/ddf2f9ade211195372236c2f39a75b00fa18d4de/src/registries/npm-registry.js#L329-L348), I really think this module should also handle it.

pvdlg:
When I say it's an edge case, I meant that you cannot obtain such result while configuring with the `npm config` command, which is the recommended approach everywhere in the npm doc. Also it doesn't seems to part of the public API as it's not documented anywhere.

For information, the only config that the `npm` CLI write as registry scope seems to be the `_authToken` which we support.

In any case, this seems to be a better fit for https://github.com/kevva/npm-conf.

We prefer to rely on external libraries, if possible from `npm` to deal with everything `npm` related, as the focus of this plugin is only to extend `semantic-release` not to reimplement/reverse engineer `npm` functionality.
If we would start to deal with `npm` undocumented internal stuff, we'll have to spend much of our time dealing with bugs due to changes in those undocumented features on `npm` side (bugs like "It works with npm version x.y.1 but breaks on npm x.y.2 in such and such situation"). 

Ideally this plugin should deal only with things that are part of the `npm` public API (documented). On a side note, that also mean writing the token in the `.npmrc` is something that should be part of the `npm` API (something like `npm publish --token $NPM_TOKEN`).

If it turns out that registry scoped parameters is a must have for some users and https://github.com/kevva/npm-conf doesn't implement it, then we will consider in this plugin to refactor the `get-last-release` part to use the `npm` CLI and call `npm view`.
We already made that choice for publishing, because https://github.com/npm/npm-registry-client implement only a small subset of what the `npm` CLI does when publishing.

pvdlg:
Please open an issue/PR for this use case on https://github.com/kevva/npm-conf.

If it gets refused there, and this feature is absolutely necessary, please re-open an issue here and we'll consider refactoring the `get-last-release` part to use the `npm` CLI and call `npm view`.

PR #26, opened 2017-12-22
Retrieve gitHead from git tag if it's missing from the npm metadata

pvdlg:
This is a port of a workaround for npm/read-package-json#77  implemented in the core when npm publishing was still part of it.

Now that we handle `npm` via plugin this npm specific workaround should be here.

In addition this will have to be used within this plugin to implement semantic-release/semantic-release#563

### DIFF
--- a/lib/get-last-release.js
+++ b/lib/get-last-release.js
@@ -4,6 +4,7 @@ const npmConf = require('npm-conf');
 const RegClient = require('npm-registry-client');
 const getClientConfig = require('./get-client-config');
 const getRegistry = require('./get-registry');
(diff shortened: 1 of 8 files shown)


pvdlg:
> Can we remove this workaround once npm/read-package-json#77 ships?

Kind of. The problem is even when it's fixed you still have a lot version already published that are missing the `gitHead`. That mean we can remove it only a while after npm/read-package-json#77 ships.

pvdlg:
> If so, maybe add a comment explaining that lib/get-version-head.js is a workaround with a reference to npm/read-package-json#77 & npm/read-package-json#76?

Done

PR #28, opened 2017-12-26
Verify `package.json` has a `version` property

### LINKED ISSUE #27: Error: package.json requires a valid "version" field
This seems to be introduces by the patch in #22. If my `package.json` doesn't contain a `version` field, I'm not able to publish it to npm.
Using a placeholder like `0.0.0-development` as I see it in this repository solves this, but is this actually the intended usage?

pvdlg:
Fix #27

### DIFF
--- a/lib/get-pkg.js
+++ b/lib/get-pkg.js
@@ -4,8 +4,13 @@ const SemanticReleaseError = require('@semantic-release/error');
 module.exports = async pkgRoot => {
   try {
     const pkg = await readPkg(pkgRoot);
+
     if (!pkg.name) {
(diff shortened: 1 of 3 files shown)


gr2m:
Thanks for the review @PostCrafter :thu

cyberhck:
first, this PR made "version" field mandatory, previously I didn't have to have version on my `package.json`, semantic-release would calculate one, and update the package.json and publish it without any hassle, my `package.json` would not contain version field, not on github, not during development, only on npm's registry, and on travis.

Second is if it's making version mandatory, shouldn't this be a breaking change? This broke my release cycle. I didn't see any change log which states that version is must have field on package.json

What's the motive behind this PR? Why does it do this? Do we really want to verify that version actually exists on package.json?

Issue #38, opened 2018-01-11
ENONPMTOKEN When npmPublish is false


piercus:
Config is 

```json
  "release": {
    "verifyConditions": ["@semantic-release/git", "@semantic-release/github"],
    "getLastRelease": "@semantic-release/git",
    "publish": [{
        "path": "@semantic-release/npm",
        "npmPublish": false
      }, 
      "@semantic-release/git", 
      "@semantic-release/github"
    ]
  }
```

Error is raised 
```bash
[Semantic release]: Call plugin publish
[Semantic release]: Verify authentication for registry https://registry.npmjs.org/
[Semantic release]: ENONPMTOKEN No npm token specified.
```

On my point of view, if `npmPublish` is set to false, there is not need to authentificate on npm. Am i missing something ?

PR #39, opened 2018-01-11
do not raise ENONPMTOKEN on publish when npmPublish is false

piercus:
Proposal in order to solve #38

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,8 +61,10 @@ async function publish(pluginConfig, {nextRelease: {version}, logger}) {
   const pkg = await getPkg(pluginConfig.pkgRoot);
   if (!verified) {
     await verifyNpmConfig(pluginConfig, pkg, logger);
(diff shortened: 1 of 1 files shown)


pvdlg:
Oops, I just saw this PR, after pushing #40...

The PR in #40 is more or less the same idea. It just modify an integration test to make sure it works as intended and still set the `verified` variable to `true` when when skipping the auth verification.

piercus:
@pvdlg I would have been happy to contribute, but seems this project is too reactive for me ;-)

And you're right, it's better with a unit test :-) (not sure about `verified`, not a big deal anyway) closing this one then.

Thank for your reactivity anyway

pvdlg:
Sorry for that. There is a lot of issues across the repositories with the "help wanted" label if you are interested. Any help is much welcome!

PR #41, opened 2018-01-11
fix: verify authentication for default npm registry only

pvdlg:
As reported on Gitter the Nexus registry doesn't support the `npm whoami` command, so the validation would always fails even when the token is valid.

After an in depth with @kbrandwijk it turns out there is no unique way to verify the token validate that would work on every custom npm registry. As npm doesn't provide any spec each custom solution made their own decisions  and we ends up with wildly different behavior:
- Not all `npm` CLI are supported (including `npm login` and `npm whoami` that are available only on a few implementation)
- The Http API doesn't return the same error code when authentication is invalid and the package has never been published. Some return a 401 some other a 404.

In addition there is no way to determine what is the implementation of the registry, so we can't adapt the behavior of our code to each one of them.

Verifying the validity of the token is still a valuable feature, but it's impossible to do it for custom implementation. This PR limit the verification of the token to the cases that use the default npm registry.


pvdlg:
> I wonder if npm registry alternatives have these differences

Mostly because there is no standard nor documentation of a registry API.

> maybe the right approach would be to have community plugins for these alternative registries that could base on @semantic-release/npm but have custom implementations for verification and publish?

Yes. But the problem is that with some implementation there is really no way to verify the token (other than actually publishing).

PR #49, opened 2018-02-06
feat: return release informations from `publish` hook

pvdlg:
Ref semantic-release/semantic-release#637


pvdlg:
Thanks! I'll keep it open for now until semantic-release/semantic-release#637 is finalized. So I can test everything.

PR #50, opened 2018-02-07
fix: get tarball name as last line from stdout

Drapegnik:
some packages, for example [`patch-package`](https://github.com/ds300/patch-package), can print output while `prepare` task, so in order to get right tarball name it's better to get only last line from stdout

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -11,7 +11,7 @@ module.exports = async ({npmPublish, tarballDir, pkgRoot}, {publishConfig, name}
 
   if (tarballDir) {
     logger.log('Creating npm package version %s', version);
-    const tarball = await execa.stdout('npm', ['pack', `./${basePath}`]);
+    const tarball = (await execa.stdout('npm', ['pack', `./${basePath}`])).split('\n').pop();
     await move(tarball, path.join(tarballDir.trim(), tarball));
   }
 


lib/publish.js
```diff
@@ -11,7 +11,10 @@ module.exports = async ({npmPublish, tarballDir, pkgRoot}, {publishConfig, name}
 
   if (tarballDir) {
     logger.log('Creating npm package version %s', version);
-    const tarball = await execa.stdout('npm', ['pack', `./${basePath}`]);
+    const tarball = (await execa.stdout('npm', ['pack', `./${basePath}`]))
+      .split('\n')
+      .slice(-1)
```
pvdlg: Why using `slice(-1)` and `pop()` here? They seems to do the same thing.
I think `slice(-1)` can be removed, so we keep only `pop()`.

Drapegnik:
Fixed.

gr2m:
@Drapegnik could you please explain how I can reproduce the problem with the additional print output? 

I created a minimal setup like so

```
mkdir test
cd test
npm init -y
npm i -D patch-package
npm pack .
```

and the output was just

```
funk-1.0.0.tgz
```

No additional lines. Have I missed something?

gr2m:
Ah I got it now, running `npm pack .` in the https://github.com/ds300/patch-package module shows additional logs. Thanks!

PR #62, opened 2018-04-19
feat: Add support for registry configuration

### LINKED ISSUE #61: Define registry and tag into plugin config?
I'm using `semantic-release-monorepo` and I'd like to share the same `publishConfig` across all my packages.

Right now I have to define:

```
"publishConfig": {
  "registry": "my custom registry",
  "tag": "latest"
}
```

on each package.
It would be useful to have the ability to define the `registry` and `tag` properties in the plugin config so that they will be shared by all the packages.

Example:

```
  "release": {
    "publish": [
      {
        "path": "@semantic-release/npm",
        "registry": "registry here",
        "tag": "latest"
      }
    ],
  }
```

FezVrasta:
With this change an user is going to be able to define a custom registry for npm without having to edit its `package.json`.

This is especially useful in conjunction with `semantic-release-monorepo`, because it allows to define the custom registry in the root configuration and have it applied to all the monorepo packages.

Fixes #61

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -2,10 +2,10 @@ const execa = require('execa');
 const getRegistry = require('./get-registry');
 const getReleaseInfo = require('./get-release-info');
 
-module.exports = async ({npmPublish, pkgRoot}, {publishConfig, name}, version, logger) => {
+module.exports = async ({npmPublish, pkgRoot, registry}, {publishConfig, name}, version, logger) => {
   if (npmPublish !== false) {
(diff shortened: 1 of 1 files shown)


pvdlg:
See https://github.com/semantic-release/npm/issues/61#issuecomment-382900376

I'm going to close this PR as using an `.npmrc` file is the recommended approach and there is no plans to accommodate to `semantic-release-monorepo`.

PR #81, opened 2018-06-27
fix: validate only known options

pvdlg:


pvdlg:
Yep I made in a mistake in the previous one :-0

PR #82, opened 2018-06-27
fix: don't attempt to call an undefined validator

kentcdodds:
I'm not sure in what situation this manifests, but my releases have been failing since this was released: https://travis-ci.org/kentcdodds/babel-plugin-preval/builds/397549428


pvdlg:
I unpublished the version `v3.3.3` that contains this bug and I fixed it in #81.
It is released in `v3.3.4`.
Sorry for the inconvenience...

kentcdodds:
Perfect. Thank you!

PR #83, opened 2018-06-29
fix: fixes incorrect escaping on legacy workflow

mcjfunk:
The auth string that is written to the .npmrc file during the legacy workflow contained variable
names rather than values. Such as:

```
// .npmrc
_auth = ${LEGACY_TOKEN}
email = ${NPM_EMAIL}
```
 This commit fixes incorrect escaping to resolve this.

### DIFF
--- a/lib/set-npmrc-auth.js
+++ b/lib/set-npmrc-auth.js
@@ -6,13 +6,13 @@ const getError = require('./get-error');
 
 module.exports = async (registry, logger) => {
   logger.log('Verify authentication for registry %s', registry);
-  const {NPM_TOKEN, NPM_USERNAME, NPM_PASSWORD, NPM_EMAIL} = process.env;
+  const {NPM_TOKEN, NPM_USERNAME, NPM_PASSWORD, NPM_EMAIL, LEGACY_TOKEN} = process.env;
 
   if (getAuthToken(registry)) {
     return;
   }
   if (NPM_USERNAME && NPM_PASSWORD && NPM_EMAIL) {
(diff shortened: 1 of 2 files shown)


pvdlg:
The escaping is on purpose so we don't write the token value in a file for securities reasons.
The `LEGACY_TOKEN` environment variable is defined in [set-npmrc-auth.js](https://github.com/semantic-release/npm/blob/master/lib/set-npmrc-auth.js) and is resolved by npm when running a command. So the code works as expected.

What issue are you trying to solve?

mcjfunk:
Good to know. It seemed purposeful. 

I'm attempting to set up semantic-release in Jenkins with Artifactory as the npm repository. After semantic-releases runs, I see this:
```
01:09:29 Error: Failed to replace env in config: ${LEGACY_TOKEN}
01:09:29     at /opt/nvm/versions/node/v8.11.0/lib/node_modules/npm/lib/config/core.js:417:13
01:09:29     at String.replace (<anonymous>)
01:09:29     at envReplace (/opt/nvm/versions/node/v8.11.0/lib/node_modules/npm/lib/config/core.js:413:12)
```
I think I may have to delete the .npmrc file after semantic-release does its work.

pvdlg:
We didn't plan that one would use npm after semantic-release had run.
What's are you using npm for after semantic-release had run?

mcjfunk:
Ultimately, I don't think I'll need to run npm after semantic-release. I'm working towards that state. 

I'm good now, closing this PR. Thanks for the wonderful library.

PR #118, opened 2018-11-23
fix for recent change of NPM publishing policy

smnbbrv:
Since a couple of days NPM broke the normal publishing process.

image

The problem and solution are present here:

https://stackoverflow.com/questions/53420758/npm-publish-gives-unscoped-packages-cannot-be-private

In fact, the change is that now in order to push the public package one must set the access parameter to public:

```
npm publish --access public
```

This PR aims to keep the original NPM behavior and sets the default value to 'restricted', however there must be a way to set it to public when one needs it (say every OSS project will suffer otherwise)

Please review ASAP.


smnbbrv:
Updated tests as well, seems the `restricted` needs to be additionally tested...

pvdlg:
As far as I know it's not new, unscoped package always require the `access` to be `public` for the first publish.

There is already an npm option for that, that you can [set in your `package.json`](https://docs.npmjs.com/files/package.json#publishconfig):
```json
"publishConfig": {
  "access": "public"
 }
```

smnbbrv:
@pvdlg thanks for the comment and help! Using `publishConfig` works.

However the NPM change seems to be quite recent. The stackoverflow question got more than 60 upvotes within 3 days. Probably it still worth to mention your solution in the README, the people will ask this again sooner or later...

PR #161, opened 2019-05-14
registry option

### LINKED ISSUE #147: Private registry as an environment variable or plugin config
Hello,

The native NPM api uses the `.npmrc` file or `publishConfig` object from the `package.json` to set publishing configuration (registry, authentication...). I somehow disagree with this setup as I believe deployment configuration shouldn't be embedded in project files, but set at runtime in the CI environment.

You have done a great job at respecting NPM's API. I was wondering if it would be possible to be able to set the NPM `registry` directly in the plugin configuration or via an environment variable and use it in the `--registry` option of most NPM commands. 

What is your take on this ? Would you consider it as an additional feature ?

All the best and thanks for the great plugin !

gr2m:
closes #147

I’ve published a fork for testing `npm install -D @gr2m/semantic-release-npm-161`, then configure the fork to be used

```json
  "semantic-release": {
    "plugins": [
      "@semantic-release/commit-analyzer",
      "@semantic-release/release-notes-generator",
      [
        "@gr2m/semantic-release-npm-161",
        {
          "registry": "https://npm.pkg.github.com/"
        }
      ],
      "@semantic-release/github"
    ]
  },
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,8 +85,7 @@
     "trailingComma": "es5"
   },
   "publishConfig": {
-    "access": "public",
-    "tag": "next"
+    "access": "public"
   },
   "repository": {
     "type": "git",
--- a/README.md
+++ b/README.md
@@ -124,6 +124,24 @@ Publish to [GitHub’s npm registry](https://help.github.com/en/articles/configu
 }
 ```
 
+Publish to multiple registries
+
+```json
+{
+  "plugins": [
+    "@semantic-release/commit-analyzer",
+    "@semantic-release/release-notes-generator",
+    [
+      "@semantic-release/npm",
+      {
+        "registry": ["https://npm.pkg.github.com/", "https://registry.npmjs.org/"]
+      }
+    ],
+    "@semantic-release/github"
+  ]
+}
+```
+
 When publishing from a sub-directory with the `pkgRoot` option, the `package.json` and `npm-shrinkwrap.json` updated with the new version can be moved to another directory with a `postpublish` [npm script](https://docs.npmjs.com/misc/scripts). For example with the [@semantic-release/git](https://github.com/semantic-release/git) plugin:
 
 ```json
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 [![npm next version](https://img.shields.io/npm/v/@semantic-release/npm/next.svg)](https://www.npmjs.com/package/@semantic-release/npm)
 
 | Step               | Description                                                                                                                                   |
-|--------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|
+| ------------------ | --------------------------------------------------------------------------------------------------------------------------------------------- |
 | `verifyConditions` | Verify the presence of the `NPM_TOKEN` environment variable, create or update the `.npmrc` file with the token and verify the token is valid. |
 | `prepare`          | Update the `package.json` version and [create](https://docs.npmjs.com/cli/pack) the npm package tarball.                                      |
 | `publish`          | [Publish the npm package](https://docs.npmjs.com/cli/publish) to the registry.                                                                |
@@ -27,11 +27,7 @@ The plugin can be configured in the [**semantic-release** configuration file](ht
 
 ```json
 {
-  "plugins": [
-    "@semantic-release/commit-analyzer",
-    "@semantic-release/release-notes-generator",
-    "@semantic-release/npm",
-  ]
+  "plugins": ["@semantic-release/commit-analyzer", "@semantic-release/release-notes-generator", "@semantic-release/npm"]
 }
 ```
 
@@ -59,9 +55,10 @@ Use either `NPM_TOKEN` for token authentication or `NPM_USERNAME`, `NPM_PASSWORD
 ### Options
 
 | Options      | Description                                                                                                         | Default                                                                                                                          |
-|--------------|---------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------|
+| ------------ | ------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
 | `npmPublish` | Whether to publish the `npm` package to the registry. If `false` the `package.json` version will still be updated.  | `false` if the `package.json` [private](https://docs.npmjs.com/files/package.json#private) property is `true`, `true` otherwise. |
 | `pkgRoot`    | Directory path to publish.                                                                                          | `.`                                                                                                                              |
+| `registry`   | Registry to publish to.                                                                                             | `https://registry.npmjs.org/`                                                                                                    |
 | `tarballDir` | Directory path in which to write the the package tarball. If `false` the tarball is not be kept on the file system. | `false`                                                                                                                          |
 
 **Note**: The `pkgRoot` directory must contains a `package.json`. The version will be updated only in the `package.json` and `npm-shrinkwrap.json` within the `pkgRoot` directory.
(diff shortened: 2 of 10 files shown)


pvdlg:
What's the reason to add an option to the plugin rather than using `.npmrc` or the `registry` field in `package.json`?

The general approach was to always use npm configuration ability (.npmrc file) when possible and rely to plugin option only when there is no other choices.
If we add options to the plugin that can also be configured via `.npmrc` we open the door to complex and hard to debug situation in the future where the actual config come partially from `.npmrc` and partially from the plugin config.

gr2m:
> What's the reason to add an option to the plugin rather than using `.npmrc` or the `registry` field in `package.json`?

I want to be able to publish to multiple registries (in my case npm inc’s registry and GitHub’s npm registry)

pvdlg:
> I want to be able to publish to multiple registries (in my case npm inc’s registry and GitHub’s npm registry)

So shouldn't the `registry` option be an Array. This way the plugin would run the `npm publish` command for each registry.
I guessing the other solution would be to define the `@semantic-release/npm` plugin multiple times (once for each registry). But that seems cleaner to have an array and have the plugin to handle that.

gr2m:
> I guessing the other solution would be to define the `@semantic-release/npm` plugin multiple times (once for each registry). 

That was my initial idea

> So shouldn't the `registry` option be an Array. This way the plugin would run the `npm publish`command for each registry.

Yes, allowing registry to be an array sounds like a good idea

pvdlg:
> If that’d be okay with you, I can work on adding that

Sure!

package.json
```diff
@@ -85,8 +85,7 @@
     "trailingComma": "es5"
   },
   "publishConfig": {
-    "access": "public",
-    "tag": "next"
```
gr2m: oops shall I revert this change? I think we discussed that we want to remove it?

pvdlg: Yes. You should revert it and it will publish on `@next` then you can use `npm dist-tag` to move it to `@latest`

gr2m: Ah yes I remember. I feel I’ll forget it though, I’d rather publish to latest and if something goes wrong set the latest dist-tag to the previous version and fix it with another release ...

gr2m:
I think this is more complicated than I thought. `https://registry.npmjs.org` is hardcoded at several places, so is the check for `NPM_TOKEN` (or `NPM_USERNAME && NPM_PASSWORD && NPM_EMAIL`) in `/lib/set-npnrc-auth.js`. Other registries might need other authentication, but maybe it’s safe to assume they all will work with a token? Should we make the name of the environment variable configurable as well? For the GitHub registry I could set it to `GITHUB_TOKEN` and it should just work.

I wonder if this is getting to complicated and should instead be implemented in a separate plugin: https://github.com/semantic-release/github-npm-registry? The two would share a lot of logic though.

What do you think @pvdlg?

gr2m:
Just as a reminder, @charliebravodev mentioned a different motivation for adding a registry option in https://github.com/semantic-release/npm/issues/147

> The native NPM api uses the `.npmrc` file or `publishConfig` object from the `package.json` to set publishing configuration (registry, authentication...). I somehow disagree with this setup as I believe deployment configuration shouldn't be embedded in project files, but set at runtime in the CI environment.

pvdlg:
> Other registries might need other authentication, but maybe it’s safe to assume they all will work with a token? Should we make the name of the environment variable configurable as well?

This plugin is meant to publish on npm or npm compliant repo. Is the GitHub registry is that it is npm compliant (like Artifactory or other)? If yes we can support it in this plugin.
In such case we should keep `https://registry.npmjs.org` as the default and we should keep the environment variable as is. That mean for the end user, they would have to set `NPM_TOKEN` with whatever token they get from GitHub, which might or might not be the same as the one they use for `@semantic-release/github` which they would set in `GH_TOKEN`.

> The native NPM api uses the .npmrc file or publishConfig object from the package.json to set publishing configuration (registry, authentication...). I somehow disagree with this setup as I believe deployment configuration shouldn't be embedded in project files, but set at runtime in the CI environment.

Actually this is a good argument against the `registry` option.
I also think that information like the registry are environment related and not project related.
The `.npmrc` file allow you to define config at the environment level by creating it in the home directory of the CI user. This is the approach I would recommend.

The more I think about it the less I'm convinced we should add the `registry` option. I don't think publishing on multiple repository is a good thing. You need to have a single source of truth for a release. If the same release is published on both GitHub and npm registry, which one is the good one? They should theoretically but identical, but there is no way to guarantee that.

If think the GitHub registry act as a proxy of the npm one. That mean:
- If you publish on npm, user using the GitHub registry will still have access to your package
- If you publish on GitHub registry, then only users of the GitHub will have access to it (as I'm guessing GitHub proxy npm but npm doesn't proxy GitHub)

So I would recommend to:
- Not add the `registry` option
- Add doc to explain how to create an `.npmrc` (preferable in the home directory) to set the registry
- Make that GitHub does proxy npm and add doc to explain users that they can publish to GitHub registry, but if they do so only users using the GitHub registry will get their package

gr2m:
> I don't think publishing on multiple repository is a good thing

Yes I actually agree. I at least wanted to publish the `@octokit/` packages to GitHub’s registry more for testing purposes than anything else.

Another reason might be in future that people wish to transition from one registry to another, they might want to start publishing their packages to GitHub’s registry as a means of backup. But I’d say that’s a problem when it’s a problem.

I'm okay with closing this PR and add an issue to document how to use this plugin to publish to GitHub’s registry by using the `.npmrc` file or the `"registry"` key in package.json

Just to answer your other questions

> This plugin is meant to publish on npm or npm compliant repo. Is the GitHub registry is that it is npm compliant (like Artifactory or other)? If yes we can support it in this plugin.

Yes it is. Or at least it aims to be. For example the `whoami` endpoint was missing in the beginning but was recently added. If we run into a missing endpoint I’m positive it will be addressed by GitHub

> That mean for the end user, they would have to set `NPM_TOKEN` with whatever token they get from GitHub, which might or might not be the same as the one they use for `@semantic-release/github` which they would set in `GH_TOKEN`.

The problem is that if I want to publish to multiple registries and both require different tokens, than I cannot set the `NPM_TOKEN` environment variable, because it would work for one but not the other.

pvdlg:
> I'm okay with closing this PR and add an issue to document how to use this plugin to publish to GitHub’s registry by using the .npmrc file or the "registry" key in package.json

👍I think it should be up to external services to handle the problematic of multiple registry.
For example a service like [unpkg](https://unpkg.com/) could be configured a your registry and in the background it would figured out if a package is published on GitHub, npm or somewhere else and then pull it and deliver it.
The easiest solution is probably for every registry that is not npm to proxy the npm registry.
If the GitHub registry gets more users than I guess it would be the other way around and npm would start to proxy Github. 

> The problem is that if I want to publish to multiple registries and both require different tokens, than I cannot set the NPM_TOKEN environment variable, because it would work for one but not the other.

I didn't think of that. Indeed that would add a lot of complexity.

travi:
> in the background it would figured out if a package is published on GitHub, npm or somewhere else and then pull it and deliver it

it will be quite interesting to see how this part plays out where scopes collide. i imagine it will be quite common for npm and github account IDs to match. if an `.npmrc` is not involved to define how to resolve the conflict, i don't know that there is a very good answer.

travi:
actually, it would get even more interesting where they don't match if there is any type of automatic resolution since someone could potentially create a matching id where it didnt exist before in order to inject a malicious version into the supply chain.

pvdlg:
Closing per https://github.com/semantic-release/npm/pull/161#issuecomment-498853443

PR #163, opened 2019-05-22
Honor NPM_CONFIG_USERCONFIG setting

ryaninvents:
This bug cost me weeks 😅

`npm` honors the `NPM_CONFIG_USERCONFIG` setting, which our CI relies on to inject shared credentials at runtime. `npm publish` succeeds if we're not using Semantic Release. However, `@semantic-release/npm` has reimplemented a credentials check in a manner not compatible with the way `npm` handles the user config environment variable. This PR checks the location specified in `NPM_CONFIG_USERCONFIG` before using `rc` to crawl up the file hierarchy.

It should be noted that `rc` does not fully implement the `.npmrc` resolution algorithm, skipping the ability to specify config files from an env var. That's why we're facing this problem.

I've tested this locally, and this small fix would allow us to use Semantic Release 🎉

### DIFF
--- a/lib/set-npmrc-auth.js
+++ b/lib/set-npmrc-auth.js
@@ -6,9 +6,12 @@ const nerfDart = require('nerf-dart');
 const AggregateError = require('aggregate-error');
 const getError = require('./get-error');
 
-module.exports = async (registry, {cwd, env: {NPM_TOKEN, NPM_USERNAME, NPM_PASSWORD, NPM_EMAIL}, logger}) => {
+module.exports = async (
+  registry,
+  {cwd, env: {NPM_TOKEN, NPM_CONFIG_USERCONFIG, NPM_USERNAME, NPM_PASSWORD, NPM_EMAIL}, logger}
+) => {
   logger.log('Verify authentication for registry %s', registry);
-  const config = path.resolve(cwd, '.npmrc');
+  const config = NPM_CONFIG_USERCONFIG || path.resolve(cwd, '.npmrc');
   if (getAuthToken(registry, {npmrc: rc('npm', {registry: 'https://registry.npmjs.org/'}, {config})})) {
     return;
   }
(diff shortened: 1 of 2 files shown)


gr2m:
Thanks Ryan! Could you see if you can get the tests passed?

ryaninvents:
@gr2m Thanks for the nudge! Fixed linting and also added a test for this case.

I force-pushed the branch because I realized I didn't write a semantic commit message; let me know if that's an issue and I can re-create the PR.

ryaninvents:
Thanks for the merge!

havenchyk:
@gr2m @ryaninvents don't you guys think that `.npmrc` should be used if exists and only then if `.npmrc` is missed, rely on `$NPM_CONFIG_USERCONFIG`?

ryaninvents:
It's the other way around. npm checks for `$NPM_CONFIG_USERCONFIG` first and then defaults to `~/.npmrc`, so in order to maintain compatibility we should do it in that order too. The `$NPM_CONFIG_USERCONFIG` variable is only set on systems where users specifically want to override. Normally, it's blank and you end up using `~/.npmrc` anyways.

havenchyk:
@ryaninvents you're right, but we're talking not about `~/.npmrc`, but about `path.resolve(cwd, '.npmrc')`, and it's going to be a `.npmrc` - local per project config.

https://docs.npmjs.com/files/npmrc#per-project-config-file


> Each of these files is loaded, and config options are resolved in priority order. For example, a setting in the userconfig file would override the setting in the globalconfig file.

ryaninvents:
Oh, I see -- that's a good point. The `rc` package will traverse from the current directory regardless because it uses `process.cwd()` to start searching. [This sandbox](https://codesandbox.io/s/goofy-williamson-3uwpd) demonstrates reading from a custom location as well as project-local `.npmrc`.

PR #173, opened 2019-06-26
fix: revert to execa `^1.0.0`

### LINKED ISSUE #172: new dependency to execa@2 breaks backward compatibility!
the new dependency to execa@2 in 15.1.12 implies a new node version restriction to node 8.
the backward compatibility to version 5.1.11 is broken.

pvdlg:
closes #172


gr2m:
We could update our CI on Travis to use the minimal node version required so that this won’t happen again

gr2m:
I’ll make a separate PR

PR #174, opened 2019-06-26
Test in lowest suported node version

gr2m:
follow up for #172 #173

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
 
 node_js:
   - 10
-  - 8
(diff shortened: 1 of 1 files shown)


gr2m:
@pvdlg That worked, shall we set it up in other @semantic-release repositories, too?

pvdlg:
Yes. We should also add Node 12.

gr2m:
Sounds good. If you don’t mind, I’ll make help-wanted issues, so we don’t forget about it and also give others the opportunity to contribute

Issue #93, opened 2018-07-30
Support npm `auth-and-writes` 2FA setting


pvdlg:
Follow-up to #92.

It should be possible to support npm [auth-and-writes](https://docs.npmjs.com/getting-started/using-two-factor-authentication#levels-of-authentication) level of two-factor-authentication by generating the OTP on the CI with a tool like [otplib](https://github.com/yeojz/otplib).

Users would have to enable npm 2FA via [command line](https://docs.npmjs.com/getting-started/using-two-factor-authentication#enabling-2fa-from-the-command-line) and on the step where the QRCode is displayed, copy the secret key displayed below it (Something like `Or enter code: XXXXXXXX....`).
The secret would have to be added on the CI as a secured environment variable (e.g. `NPM_2FA_SECRET`).

The only constraint is that if 2FA is already enabled, there is no way to retrieve the secret key. So users would have to disable 2FA and re-enable it via CLI to obtain a new secret key.

Then the plugin would generate the OTP with:
```js
const {authenticator} = require('otplib');

const otp =authenticator.generate(process.env.NPM_2FA_SECRET);
```
And call `npm publish` with the `--otp` option.

We would also need to implement #11 and make sure that if the `auth-and-writes` option is enabled the environment variable `NPM_2FA_SECRET` is set. Unfortunately it doesn't seems there is a way to verify if `NPM_2FA_SECRET` is correct, as all the `npm` command that requires an OTP are things that makes changes server side, so we can't use them to test.

travi:
this approach would concern me a bit. considering a worm like was attempted with eslint, the environment variables could be stolen just as easily as the token. while this is a creative way to make this work, i'd really hesitate to expose the 2fa secret to anything other than my 2fa app, such as google authenticator or authy because of opening up the account to further risk.

imho, i think a better approach would be to have the ability to require 2fa for `auth-and-writes`, but exclude very specific tokens, such as the ones used by `semantic-release`. since `semantic-release` is intended to be run from CI, [limiting the CIDR ranges](https://gist.github.com/travi/f91c73610fa49769d90e3ee3b66cfbee) allowed for that token should be enough to make up for that exception. obviously, npm does not support 2fa enforcement exceptions, but [i've mentioned the idea to the team](https://npm.community/t/allow-subset-of-tokens-to-bypass-2fa-requirement/536/3). maybe a few more voices around that topic would help reinforce the importance of the automated use case.

travi:
[...]

PR #176, opened 2019-06-28
Support OTP fetching from config.otpUrl

dominykas:
[Tested](https://travis-ci.org/dominykas/allow-scripts/jobs/551973725) on one of [my own packages](https://github.com/dominykas/allow-scripts/compare/f3511752d7a4e185e14725b3e42572937852889c...c6a94eafd52fcc97536f67a51e07a61597b291d8#diff-d0fd6e011e62dec6517327248cb7766bR5): works like magic!

I think I got the coverage right, but I'm still not sure how to assert that `--otp` was passed through to npm. npm will send the [OTP inside the `npm-otp` header](https://github.com/npm/npm-registry-fetch/blob/5dbd1d722b9125026eea1f0ef019119598e7bfb0/index.js#L188), but short of adding a proxy around the couch app, I'm not sure if it's even possible to observe that in any way.

See also: #93


pvdlg:
Maybe I'm missing something here but that seems to be a huge security hole...
Anyone can look at the URL you configure in `otpUrl` calls it, get your OTP and release anything with your account.

dominykas:
No, they cannot release anything, because

a) I need to press a button on my phone (although I'll admit that someone could configure a server to respond without that, however that's no worse than no 2FA at all)
b) OTP is not enough - you still need the token (so once again - no worse than 2FA)

dominykas:
Also the URL should be kept as secret as the token (i.e. in env). I can easily update this to only use `process.env.NPM_OTP_URL` if that's the preferred route?

pvdlg:
Reading through the documentation of https://github.com/nearform/optic I don't think using it require any modification in this plugin.

In your CI you could simply set `NPM_CONFIG_OTP` (which is equivalent of passing the OPT via --otp).

Because the OTP is valid only for a few seconds you could do that via a `prepublishOnly` step in your `package.json` so it would be executed just before semantic-release runs `npm publish`:
```json
{
 "scripts": {
    "prepublishOnly": "export NPM_CONFIG_OTP=$(curl -s $NPM_OTP_URL/$OTP_TOKEN)"
  }
}
```

I would rather not make any change in this plugin not recommending any particular solution in the doc out of security concern.

dominykas:
I was not aware you could set it this way - I'll try that out, thank you.

I will try this out and report back.

I have no intent of recommending any particular solution, but if this approach works - I suppose some documentation is in order - I'll see if I can find a place to PR that.

dominykas:
No luck: https://travis-ci.org/dominykas/allow-scripts/jobs/575685664#L310

I suspected this might happen, because I had tried playing around with env vars in a different context - they don't carry through, i.e. npm does not see an env vars from one of `scripts` inside another one.

I'll try writing the otp into an `.npmrc` next, but IIRC, the config for `publish` is loaded _before_ `prepublishOnly` executes, so I'm not sure that will have an effect either.

dominykas:
No luck: https://travis-ci.org/dominykas/allow-scripts/jobs/575690340#L316

It can probably be worked around by using the `exec` plugin, but it seems to me that this is a behavior that's important enough to be built-in, rather than worked around?

Would it make sense to replace `otpUrl` with an `otpCommand` to make it more flexible (with `curl -s $URL` being one way to configure the command)?

pvdlg:
In your first test the error is `The local branch master is behind the remote one, therefore a new version won't be published.` so the failure is not related to the OTP token.

dominykas:
> In your first test the error is The local branch master is behind the remote one, therefore a new version won't be published. so the failure is not related to the OTP token.

The link I have points to line 310 in the log... but the log does not contain such a line. The release job also ran 1.5 hours later than the test jobs, so I suspect for whatever reason (quite possibly by accident) I restarted the build.

I'll see if there's a way to recover the original logs, but I'm fairly certain it was an `EOTP` originally.

pvdlg:
Closing per https://github.com/semantic-release/npm/pull/234#issuecomment-565683509

Issue #215, opened 2019-11-01
NPM publish: SemanticReleaseError: Invalid npm token.


varnastadeus:
## Current behavior

Can't publish with valid token using semantic-release. Running from local machine in dry run mode using:  
`GH_TOKEN=*** NPM_TOKEN=*** npx semantic-release`.

Running `npm publish` with the same token works as expected.  
npm version: `6.11.3`
node version: `v12.12.0`

## Expected behavior

Semantic-release should be able to publish to npm successfully.
## Environment

- **semantic-release** version: 15.13.27
- CI environment: github actions
- Plugins used: 
```
"@semantic-release/commit-analyzer",
"@semantic-release/changelog",
"@semantic-release/github
"@semantic-release/npm",
@semantic-release/release-notes-generator
```
- **semantic-release** configuration: 
```
"release": {
    "branch": "master",
    "plugins": [
      "@semantic-release/commit-analyzer",
      "@semantic-release/changelog",
      "@semantic-release/github",
      [
        "@semantic-release/npm",
        {
          "pkgRoot": "dist/ng-select"
        }
      ],
      "@semantic-release/release-notes-generator"
    ]
  }
```

- Logs running in dry mode on local machine:
```
[4:12:33 PM] [semantic-release] › ℹ  Running semantic-release version 15.13.27
[4:12:33 PM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/changelog"
[4:12:33 PM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/github"
[4:12:33 PM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/npm"
[4:12:33 PM] [semantic-release] › ✔  Loaded plugin "analyzeCommits" from "@semantic-release/commit-analyzer"
[4:12:33 PM] [semantic-release] › ✔  Loaded plugin "generateNotes" from "@semantic-release/release-notes-generator"
[4:12:33 PM] [semantic-release] › ✔  Loaded plugin "prepare" from "@semantic-release/changelog"
[4:12:33 PM] [semantic-release] › ✔  Loaded plugin "prepare" from "@semantic-release/npm"
[4:12:33 PM] [semantic-release] › ✔  Loaded plugin "publish" from "@semantic-release/github"
[4:12:33 PM] [semantic-release] › ✔  Loaded plugin "publish" from "@semantic-release/npm"
[...]

PR #218, opened 2019-11-03
fix: log the output of `npm whoami` command

pvdlg:


gr2m:
What problem is this fixing?

pvdlg:
Help debug https://github.com/semantic-release/npm/issues/215#issuecomment-549107857

PR #223, opened 2019-11-21
fix: use correct peerDependencies for semantic-release

hershmire:
`semantic-release@v16-beta` isn't satisfied for current `peerDependency` range:

```
##[error]npm WARN @semantic-release/npm@6.0.0-beta.3 requires a peer of semantic-release@>=15.9.0 <17.0.0 but none is installed.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "all": true
   },
   "peerDependencies": {
-    "semantic-release": ">=15.9.0 <17.0.0"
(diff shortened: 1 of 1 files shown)


pvdlg:
Is that necessary? `16.0.0-beta` should be > 15.9.0 isn't it?

hershmire:
`>=15.9.0 <17.0.0` range will not include any `beta` versions.

PR #268, opened 2020-07-02
docs(README): fix columns of options table in readme

rxgx:
### What's changed?

In the options table, the columns weren't matching so I've updated the markdown so there are only 2 columns. Was this intended or auto-formatted this way?


gr2m:
Thanks.  lot!

PR #273, opened 2020-09-05
Updating node and npm version in package

roggervalf:
Hi, I made this PR to update the node version since when I tried the lowest version in the package, it throws an error for got dependency
```sh
error got@11.6.0: The engine "node" is incompatible with this module. Expected version ">=10.19.0". Got "10.18.1"
error Found incompatible module.
```
I also update npm dependency because the actual version has a subdependency in its treeDependency that give me a vulnerability issue detected from github.
Hope this could help.


gr2m (approved):
Thanks

PR #274, opened 2020-09-13
Trying delete npm dependency

roggervalf:
This is just a try, should we need npm as a dependency here?

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "lodash": "^4.17.15",
     "nerf-dart": "^1.0.0",
     "normalize-url": "^5.0.0",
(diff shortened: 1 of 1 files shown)


travi:
as mentioned in other issues in this repository, the npm plugin depends very directly on npm to accomplish its goals. in order to ensure that a verified version is used, it is marked as a dependency on purpose.

if depending on semantic-release as a local dependency is causing npm to be pulled in and affect your project in undesireable ways, you might want to consider running semantic-release with `npx` instead of installing as a local dependency.

roggervalf:
Thank you @travi for the quick explanation

Issue #277, opened 2020-10-08
NPM automation token


radum:
NPM introduced automation tokens https://github.blog/changelog/2020-10-02-npm-automation-tokens/

This means we can keep our 2FA for login and publish and use automation tokens to bypass 2FA for publishing.

I did that on a test project and `semantic-release/npm` fails at the verify auth step for npm.

I was wondering it this is miss configuration on my part if there is work that needs to be done to support the new tokens.

The setup is simple, I am using 2FA for both publish and login on NPM

![image](https://user-images.githubusercontent.com/46779/95470074-426b8400-0978-11eb-89b3-09cd058acd95.png)

and my test repo has enabled the automation token option from settings.

![image](https://user-images.githubusercontent.com/46779/95469963-1f40d480-0978-11eb-9076-4f43172c6cf5.png)

As for my semantic-release config

```
{
	"branches": ["+([0-9])?(.{+([0-9]),x}).x", "main", "next", "next-major", { "name": "beta", "prerelease": true }, { "name": "alpha", "prerelease": true }],
	"plugins": [
		"@semantic-release/commit-analyzer",
		"@semantic-release/release-notes-generator",
		[
			"@semantic-release/changelog",
			{
				"changelogTitle": "# Changelog\n\nAll notable changes to this project will be documented in this file. See\n[Conventional Commits](https://conventionalcommits.org) for commit guidelines."
			}
		],
		[
			"@semantic-release/npm",
			{
				"tarballDir": "pack"
			}
		],
		[
			"@semantic-release/github",
			{
				"assets": "pack/*.tgz"
			}
		],
		[
			"@semantic-release/git",
			{
				"message": "chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
			}
		]
	]
}
```

radum:
So it looks like if I use `npm whoami` (and I am successfully logged in via npm login) it works ok.

If I use the generated whoami from the lib:

```
npm whoami --userconfig /private/var/folders/jj/temppathhere/.npmrc --registry https://registry.npmjs.org
npm ERR! code E403
npm ERR! 403 403 Forbidden - GET https://registry.npmjs.org/-/whoami
npm ERR! 403 In most cases, you or one of your dependencies are requesting
npm ERR! 403 a package version that is forbidden by your security policy.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/radu.micu/.npm/_logs/2020-10-08T15_43_06_087Z-debug.log
[...]

PR #278, opened 2020-10-14
Allow use of npm automation tokens

phillipj:
Finally decided to give semantic-release a try, especially since npm recently added support for automation tokens (so I can still have 2FA enabled).

Stumbled upon a challenge when trying to use those shiny new tokens tho, and noticed a hint in https://github.com/semantic-release/npm/issues/277#issuecomment-705784318 about adding a new config option to avoid the npm whoami check when automation tokens are used.

Hence this PR; the introduction of `.usesAutomationToken` config option.

I got my very first automated release made with this in place (+ some other hacks to make the `semantic-release` package use my fork), which obviously felt awesome 🎉 

Any thoughts?

Refs https://github.com/semantic-release/npm/issues/277

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,6 +22,7 @@ async function verifyConditions(pluginConfig, context) {
     pluginConfig.npmPublish = defaultTo(pluginConfig.npmPublish, publishPlugin.npmPublish);
     pluginConfig.tarballDir = defaultTo(pluginConfig.tarballDir, publishPlugin.tarballDir);
     pluginConfig.pkgRoot = defaultTo(pluginConfig.pkgRoot, publishPlugin.pkgRoot);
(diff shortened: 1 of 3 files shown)


gr2m:
Thanks Phillip, I've contact the npm folks about the problem of verification, I'd prefer to await their response. They just shipped npm7 so they had other priorities, but I hope they get back to me soon now. If they enable a way to verify a token, adding this feature will become obsolete, so I'd prefer not to add it in the first place if we can help it

phillipj:
Appreciate the quick reply!

Sounds like a good idea. Fingers crossed this package won't have to handle automation tokens specifically..

Let's close this PR for now, and plausibly have it as a fallback for further discussion if needed.

PR #284, opened 2020-11-07
docs: Fix postpublish script when "pkgRoot":"dist"

hamzahamidi:
When "pkgRoot" is set the postpublish script will excute relativly to the pkgRoot. Hence cp dist/ will generate the error "No such file or directory"


gr2m (approved):
Thanks 👍🏼

PR #304, opened 2020-12-08
feat: Use npm v7

danez:
I updated npm from v6 to v7. Reading the blogposts there shouldn't be any breaking changes, at least not in the functionality that semantic-release uses.

One test failed, but it was testing npm internal stuff about formating of package.json, so I removed it. (npm v7 does seem to preserve whatever formatting is in the file and just replace the version.) Hope that is okay?

### DIFF
--- a/test/prepare.test.js
+++ b/test/prepare.test.js
@@ -189,32 +189,6 @@ test('Preserve indentation and newline', async (t) => {
   t.deepEqual(t.context.log.args[0], ['Write version %s to package.json in %s', '1.0.0', cwd]);
 });
 
-test('Use default indentation and newline if it cannot be detected', async (t) => {
-  const cwd = tempy.directory();
-  const npmrc = tempy.file({name: '.npmrc'});
-  const packagePath = path.resolve(cwd, 'package.json');
-  await outputFile(packagePath, `{"name": "package-name","version": "0.0.0-dev"}`);
-
-  await prepare(
-    npmrc,
-    {},
-    {
-      cwd,
-      env: {},
-      stdout: t.context.stdout,
-      stderr: t.context.stderr,
-      nextRelease: {version: '1.0.0'},
-      logger: t.context.logger,
-    }
-  );
-
-  // Verify package.json has been updated
-  t.is(await readFile(packagePath, 'utf-8'), `{\n  "name": "package-name",\n  "version": "1.0.0"\n}\n`);
-
-  // Verify the logger has been called with the version updated
-  t.deepEqual(t.context.log.args[0], ['Write version %s to package.json in %s', '1.0.0', cwd]);
-});
-
 test('Create the package in the "tarballDir" directory', async (t) => {
   const cwd = tempy.directory();
   const npmrc = tempy.file({name: '.npmrc'});
--- a/package-lock.json
+++ b/package-lock.json
@@ -623,6 +623,3537 @@
         "registry-auth-token": "^4.0.0",
         "semver": "^7.1.2",
         "tempy": "^1.0.0"
+      },
+      "dependencies": {
+        "npm": {
+          "version": "6.14.9",
+          "resolved": "https://registry.npmjs.org/npm/-/npm-6.14.9.tgz",
+          "integrity": "sha512-yHi1+i9LyAZF1gAmgyYtVk+HdABlLy94PMIDoK1TRKWvmFQAt5z3bodqVwKvzY0s6dLqQPVsRLiwhJfNtiHeCg==",
+          "dev": true,
+          "requires": {
+            "JSONStream": "^1.3.5",
+            "abbrev": "~1.1.1",
+            "ansicolors": "~0.3.2",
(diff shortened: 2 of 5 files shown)


danez:
Not sure why the integration tests fail with 403 errors.

gr2m:
Thanks a lot Daniel, I really want us to upgrade to `npm` 7, it hopefully get rid of several outdated packages that we are stuck with because of npm 6. I'll try to have a look at the failed CI. This week is busy for me because of GitHub Universe, but I'll do my best. I definitely have it on my list and won't forget

ext:
Possible a bit off-topic but I am a bit curious as to why it has to be a dependency at all, is not not possible to use system npm?

gr2m:
> not possible to use system npm?

the problem is compatibility. WE don't know what system npm version you have and making sure that the code is compatible with all current and future version is a maintenance nightmare

ext:
> > not possible to use system npm?
> 
> the problem is compatibility. WE don't know what system npm version you have and making sure that the code is compatible with all current and future version is a maintenance nightmare

But it is also a bit nightmare-ish for users? It pulls many extra dependencies (some with security vulnerabilities) and it also causes incompatibilities in run scripts, e.g. system with npm 7 uses npm 6 (or vice versa after this PR) when using `npm` in a run script because `node_modules/.bin/npm` in now in `$PATH`. (Bonus points when `node_modules/.bin/npm` says there is an update to NPM despite the system already runs the latest)

But I get your point, I was mostly curious about the reasoning.

danez:
@gr2m I figured out why the integration tests fail, It is only because npm v7 does not set the [maintainers field anymore](https://github.com/npm/libnpmpublish/commit/ecaeb0b203882f633e3e850926885eb01b2c4776#diff-5e733ad892a637d40e61205f71cd63352d98d5ddfd929ab000003eb1eee67fbcL121) on publish, but [npm-registry-couchapp](https://github.com/npm/npm-registry-couchapp/blob/5a47400a769abe07528782a8c14aca067063592f/registry/updates.js#L328) requires it to be set.

To fix this I switched to the proper npm registry [verdaccio](https://github.com/verdaccio/verdaccio) instead of the docker image from [here](https://github.com/semantic-release/npm-registry-docker).

Now all the test seems to work :)

ext (approved):

test/prepare.test.js
```diff
@@ -189,32 +189,6 @@ test('Preserve indentation and newline', async (t) => {
...
-      logger: t.context.logger,
-    }
-  );
-
-  // Verify package.json has been updated
-  t.is(await readFile(packagePath, 'utf-8'), `{\n  "name": "package-name",\n  "version": "1.0.0"\n}\n`);
-
-  // Verify the logger has been called with the version updated
-  t.deepEqual(t.context.log.args[0], ['Write version %s to package.json in %s', '1.0.0', cwd]);
-});
```
gr2m, line 216: Why did you remove this test?

gr2m (commented):
great work! Thank you so much! Just one question

test/prepare.test.js
danez: Because it was testing npm internal stuff. Npm v7 does not format `package.json` files anymore, so this test failed because no newlines were present in the expected `package.json`.
I could have "fixed" the test and adjusted the expected output, but the test before this one does exactly the same thing then and the testname would have not really described anymore what it is testing. So I decided to remove it.

gr2m: got it, thanks 👍🏼

weaintplastic:
Hey @danez. Long time no see. Hope you ar doing well.

We've updated semantic-release/npm from 7.0.10 to 7.1.0 and since then our release flow isn't able to pubish packages anymore to our private repository due to an authentification error. The release is performed inside a Github action providing an`NPM_TOKEN`

```yaml
- name: Release
  env:
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
  run: npm run release -- --debug
```

The error we're receiving is

```
npm ERR! code ENEEDAUTH 
1869npm ERR! need auth This command requires you to be logged in. 
1870npm ERR! need auth You need to authorize this machine using `npm adduser`
```

Reverting the dependency update made our releases to finish successfully again. Do you have any advice where to look for a fix?

danez:
hey @weaintplastic,

That is odd as this error only appears if no credentials are found for the registry. https://github.com/npm/cli/blob/latest/lib/publish.js#L99-L109
And semantic release should provide valid credentials for the command.

what is the npm version that is now installed in your project? 
What was the complete output of semantic release?
I guess you specify the private registry in `package.json->publishConfig->registry`?

weaintplastic:
@danez we are using node 15 and npm 7. 

we have a `.npmrc` file in our root that includes the read-only `NPM-TOKEN`.
The token with publishing rights is provided through a Github secret as posted above.
The registry is defined as you said in `package.json/publishConfig/registry`.

PR #306, opened 2020-12-10
fix(deps): updating npm dependency for CWE-400

blaky:
The previous patch version of NPM was vulnerable by CWE-400, so I am just updating it to the latest patch version that resolved the security issue.


gr2m (approved):
Thank you!

PR #362, opened 2021-04-29
fix: use `NPM_CONFIG_USERCONFIG` in get-registry to match auth

### LINKED ISSUE #349: get-registry.js doesn't take into account NPM_CONFIG_USERCONFIG
When `get-registry.js` looks for the registry it just looks at `./.npmrc` instead of taking into account `NPM_CONFIG_USERCONFIG`. This ends up requiring that you specify `NPM_CONFIG_REGISTRY` when you shouldn't have to because it's already in the alternative rc file specified by `NPM_CONFIG_USERCONFIG`.

jameshartig:
Previously NPM_CONFIG_USERCONFIG was only used for getting the auth token and not used for actually getting the registry from .npmrc.

Fixes #349

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,12 +45,13 @@ Both the [token](https://docs.npmjs.com/getting-started/working_with_tokens) and
 
 ### Environment variables
 
-| Variable       | Description                                                                                                                   |
-| -------------- | ----------------------------------------------------------------------------------------------------------------------------- |
-| `NPM_TOKEN`    | Npm token created via [npm token create](https://docs.npmjs.com/getting-started/working_with_tokens#how-to-create-new-tokens) |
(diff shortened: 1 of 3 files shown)


gr2m (commented):
Thanks for sending the pull request!

Can you please add a test to make sure we don't break this behavior in future!

rodoabad (approved):

gr2m (approved):
I confirmed locally that the test fails without the fix. Thanks for the PR @jameshartig 👍🏼

gr2m (changes requested):
oh sorry, one more thing, I think we should document the env variable in https://github.com/semantic-release/npm#environment-variables?

jameshartig:
@gr2m good call! I updated the README

gr2m (approved):

Issue #211, opened 2019-10-28
TypeError: Found non-callable @@iterator on verify conditions (take 2!)


FauxFaux:
I'm still seeing this error on the latest, previously reported as #207.

`TypeError: Found non-callable @@iterator at verifyConditions`

It when running semantic-release in a private project, and when running via `npx`.

It happens in https://github.com/FauxFaux/testcase-semantic-release , although you will need to have write access to github to get to the point, making it hard to demo in docker.

```
% npm run semantic-release 

> revise@1.0.0 semantic-release /var/tmp/faux191028.revise
> semantic-release

[1:49:39 PM] [semantic-release] › ℹ  Running semantic-release version 15.13.28
[1:49:40 PM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/npm"
[1:49:40 PM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/github"
[1:49:40 PM] [semantic-release] › ✔  Loaded plugin "analyzeCommits" from "@semantic-release/commit-analyzer"
[1:49:40 PM] [semantic-release] › ✔  Loaded plugin "generateNotes" from "@semantic-release/release-notes-generator"
[1:49:40 PM] [semantic-release] › ✔  Loaded plugin "prepare" from "@semantic-release/npm"
[1:49:40 PM] [semantic-release] › ✔  Loaded plugin "publish" from "@semantic-release/npm"
[1:49:40 PM] [semantic-release] › ✔  Loaded plugin "publish" from "@semantic-release/github"
[1:49:40 PM] [semantic-release] › ✔  Loaded plugin "success" from "@semantic-release/github"
[1:49:40 PM] [semantic-release] › ✔  Loaded plugin "fail" from "@semantic-release/github"
[1:49:40 PM] [semantic-release] › ⚠  This run was not triggered in a known CI environment, running in dry-run mode.
[1:49:40 PM] [semantic-release] › ⚠  Run automated release from branch master in dry-run mode
[1:49:43 PM] [semantic-release] › ✔  Allowed to push to the Git repository
[1:49:43 PM] [semantic-release] › ℹ  Start step "verifyConditions" of plugin "@semantic-release/npm"
[1:49:43 PM] [semantic-release] [@semantic-release/npm] › ℹ  Verify authentication for registry https://registry.npmjs.org/
[...]

Issue #207, opened 2019-10-24
TypeError: Found non-callable @@iterator on verify conditions


semoal:
As the title says, our pipeline on Gitlab stopped working. It's using node:10-alpine.

```sh
[10:37:04 AM] [semantic-release] › ✖  An error occurred while running semantic-release: { TypeError: Found non-callable @@iterator
    at verifyConditions (/builds/secret/secretlibraries/react-ui-kit/node_modules/semantic-release/node_modules/@semantic-release/npm/index.js:39:12) pluginName: '@semantic-release/npm' }
{ AggregateError: 
    TypeError: Found non-callable @@iterator
        at verifyConditions (/builds/secret/secretlibraries/react-ui-kit/node_modules/semantic-release/node_modules/@semantic-release/npm/index.js:39:12)
    at /builds/secret/secretlibraries/react-ui-kit/node_modules/semantic-release/lib/plugins/pipeline.js:54:11 name: 'AggregateError' }npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @secretbank/react-ui-kit@2.12.0-canary.38 semantic-release: `npx semantic-release --debug`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @secretbank/react-ui-kit@2.12.0-canary.38 semantic-release script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
```

Versions of semantic-release/npm tried: ["5.2.0-beta.6", "^6.0.0-beta.1", "5.3.0", "5.2.0"]

semoal:
Probably is not related directly to npm package, since i put all the semantic release packages on a fixed version, and the issue got fixed.

I paste the package.json..
```json
    "@semantic-release/changelog": "3.0.4",
    "@semantic-release/commit-analyzer": "7.0.0-beta.2",
    "@semantic-release/exec": "3.3.6",
    "@semantic-release/git": "7.1.0-beta.3",
    "@semantic-release/npm": "5.2.0",
    "semantic-release": "16.0.0-beta.22",
```

pvdlg:
So is your problem solved? If it is please close the issue.

semoal:
The problem is solved but i guess is some bug on the other libraries. I close anyways.

oytuntez:
We are having the same issue in multiple environments. @semoal, your versions didn't work for us.


```
[...]

PR #388, opened 2021-09-02
fix: don't spread errors blindly

Gopikrishna19:
This is to fix recurring problems with @@iterator #207 #211

### The problem

```
{ AggregateError: 
    TypeError: Found non-callable @@iterator
        at verifyConditions (/node_modules/@semantic-release/npm/index.js:39:12)
        at /node_modules/semantic-release/lib/plugins/pipeline.js:54:11 name: 'AggregateError' }
```
```
{ AggregateError: 
    TypeError: Found non-callable @@iterator
        at verifyConditions (/node_modules/@semantic-release/npm/index.js:38:12)
        at /node_modules/semantic-release/lib/plugins/pipeline.js:54:11 name: 'AggregateError' }
```

The errors refer to the following lines where we are spreading the error. If there is another error that is not thrown by an underlying utility function that is not an array then, spreading that causes the `non-callable @@iterable` error.

https://github.com/semantic-release/npm/blob/b26b6e9f8f9ec384a6332e4e623ec6ce6ec9b15f/index.js#L38-L40

### The fix

Do not spread the error blindly, rather concatenate it to the errors list.


gr2m:
Can you add some more context on what the problem is, and how your code fixes it? Can you add tests to avoid future regressions?

Gopikrishna19:
@gr2m the two issues I linked about gives a detailed explanation of what the problem is. The tests existing tests should not break as the behaviour did not change, merely a refactor.

```js
errors.push(...error); // fails if error is not iterable

errors = errors.concat(error); // works for both array as well as non iterable entities

// example non-iterable
[1, 2, 3].concat(4) === [1, 2, 3, 4];  // true

// example iterable
[1, 2, 3].concat([4, 5]) === [1, 2, 3, 4, 5];  // true
```

Hope that helps. There is no behaviour change and there shouldn't be any tests breaking.

Gopikrishna19:
Thinking again, I can add a test against throwing a non-iterable object. I am not exactly familiar with the testing style you guys have here. I would be happy to add the tests if you guys can help me a bit on where to put and how to mock. Generally, in `jest`, I'd do something like (pseudo):

```js
jest.mock(verifyNpmConfig).mockRejectedValues(nonIterableErrorThatHappenedForSomeOtherReason);

expect(() => verifyConditions()).toThrow([nonIterableErrorThatHappenedForSomeOtherReason]); 
                                         // to be aggregated without throwing @@iterable error
```

gr2m:
okay I got it now, cheers. However it seems that the code change did in fact change some behavior, can you look into the failing tests?

Gopikrishna19:
It's not passing in Windows or Windows WSL. Even master is not working. Is there a specific setup I have to do to replicate the testing environment? 

Theoretically speaking it shouldn't break at all 🤔 

I guess you guys are mutating the original error array by using push and the tests are based on that. If I can't replicate the environment I shall try reverting the command to push but use an array check before spreading. Pushing in a few mins.

Gopikrishna19:
@gr2m please approve for tests now. I wasn't able to test locally as Windows as the testing environment is not working.

Gopikrishna19:
I can't build it locally to test the PR and see why it is failing. Closing it. Hope someone else with a testable env takes this work. Or I can reopen the PR if/when authors support platform-independent development. Thank you for the opportunity!

PR #413, opened 2021-10-19
fix(README): correct docs on authentication

swain:
## Motivation
The documentation about authentication strategies seems to be incorrect.

The documentation implies that an `NPM_TOKEN` environment variable will take precedence over (and replace any token within) an `.npmrc` value:

> Verify the presence of the NPM_TOKEN environment variable, create or update the .npmrc file with the token and verify the token is valid.

Based on the code [here](https://github.com/semantic-release/npm/blob/master/lib/set-npmrc-auth.js#L27-L46), that implication is not correct -- `.npmrc` files take precedence over both `NPM_TOKEN` and the legacy env variables. In particular, the `return` statement on line 29 is causing the environment variables to be ignored.

## Changes
- Update the README to clearly communicate that `.npmrc` files will take precedence over the `NPM_TOKEN` environment variable.
- Added a test illustrating the behavior.

### DIFF
--- a/test/set-npmrc-auth.test.js
+++ b/test/set-npmrc-auth.test.js
@@ -212,3 +212,30 @@ test.serial('Throw error if "NPM_EMAIL" is missing', async (t) => {
   t.is(error.message, 'No npm token specified.');
   t.is(error.code, 'ENONPMTOKEN');
 });
+
+test.serial('Prefer .npmrc over environment variables', async (t) => {
+  process.env.HOME = tempy.directory();
+  const cwd = tempy.directory();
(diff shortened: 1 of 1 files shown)


gr2m:
Did you verify this behavior or is it just based on your assessment of the code? I do agree with your assessment but just want to be sure. Do you think we could add a test to confirm your assumption? I couldn't find one which tests the behavior if both an auth token is set in `.npmrc` and `NPM_TOKEN` is set, I think it would need to go into https://github.com/semantic-release/npm/blob/master/test/set-npmrc-auth.test.js

PR #416, opened 2021-10-27
fix: avoid multiple `verifyConditions` invocation

### LINKED ISSUE #414: Broken `verify-auth` implementation
**TL;DR**: I receive `ENEEDAUTH` npm errors until I comment out [these three lines](https://github.com/semantic-release/npm/blob/master/index.js#L42-L44) in my node_modules directory -- then publication works fine.

### What I did

I attempted to use [multi-semantic-release](https://github.com/dhoulb/multi-semantic-release) to semantically-publish the [typescript-tools](https://github.com/typescript-tools/typescript-tools/) monorepo. I configured the `NPM_TOKEN` environment variable with an Automation-type npm token

Both locally and in GitHub actions I invoked

```
npx multi-semantic-release
```

### What I expected to happen

I expected packages to publish semantically.

### What actually happened

I received the npm error

```
npm ERR! code ENEEDAUTH
npm ERR! need auth This command requires you to be logged in.
npm ERR! need auth You need to authorize this machine using `npm adduser`
```

### More details

Note: I originally opened [an issue](https://github.com/dhoulb/multi-semantic-release/issues/82) in multi-semantic-release but determined the root cause did not lie with code in that repository.

I have 24 packages in the typescript-tools monorepo. While instrumenting the @semantic-release/npm code with debug statements I learned that some (but not all!) of the `npm whoami` calls in [this try block](https://github.com/semantic-release/npm/blob/7338fc25de82058ea83b1e6e6669707036c2d96a/lib/verify-auth.js#L21) were failing. Plenty of them succeeded in printing my npm username in my stdout.

I was sure from my debugging that the npm credentials were being configured in the temporary `.npmrc` file correctly, so I tried commenting out the [`throw` statement handling the `whoami` errors](https://github.com/semantic-release/npm/blob/master/index.js#L42-L44) and sure enough, I was able to publish!

```
[6:56:59 PM] › 🎉  Released 5 of 24 packages, semantically!
```

So something must be going wrong with the `whoami` check. I don't think it's being rate-limited; I tried making 100 parallel calls to `npm whoami` using the @semantic-release/npm temporary `.npmrc` configuration and all 100 calls succeeded.

antongolub:
closes #414

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,7 +32,7 @@ async function verifyConditions(pluginConfig, context) {
     const pkg = await getPkg(pluginConfig, context);
 
     // Verify the npm authentication only if `npmPublish` is not false and `pkg.private` is not `true`
-    if (!verified && pluginConfig.npmPublish !== false && pkg.private !== true) {
+    if (pluginConfig.npmPublish !== false && pkg.private !== true) {
       await verifyNpmAuth(npmrc, pkg, context);
     }
   } catch (error) {
--- a/lib/verify-auth.js
+++ b/lib/verify-auth.js
@@ -5,6 +5,8 @@ const getError = require('./get-error');
 const getRegistry = require('./get-registry');
 const setNpmrcAuth = require('./set-npmrc-auth');
 
+const memo = {};
+
 module.exports = async (npmrc, pkg, context) => {
   const {
     cwd,
@@ -17,12 +19,20 @@ module.exports = async (npmrc, pkg, context) => {
   await setNpmrcAuth(npmrc, registry, context);
 
   if (normalizeUrl(registry) === normalizeUrl(DEFAULT_NPM_REGISTRY)) {
+    const key = npmrc + registry;
+    if (memo[key]) {
+      return memo[key];
+    }
+
     try {
       const whoamiResult = execa('npm', ['whoami', '--userconfig', npmrc, '--registry', registry], {cwd, env});
       whoamiResult.stdout.pipe(stdout, {end: false});
       whoamiResult.stderr.pipe(stderr, {end: false});
+
+      memo[key] = whoamiResult;
       await whoamiResult;
     } catch {
+      memo[key] = undefined;
       throw new AggregateError([getError('EINVALIDNPMTOKEN', {registry})]);
     }
   }
--- a/test/verify-auth.test.js
+++ b/test/verify-auth.test.js
@@ -1,0 +1,43 @@
+const test = require('ava');
+const {stub} = require('sinon');
+const tempy = require('tempy');
+
+const getRegistryPath = require.resolve('../lib/get-registry');
+const verifyAuthPath = require.resolve('../lib/verify-auth');
+const setNmprcAuthPath = require.resolve('../lib/set-npmrc-auth');
+const execaPath = require.resolve('execa');
+
+const resetModuleCache = () => {
+  require.cache[getRegistryPath] = undefined;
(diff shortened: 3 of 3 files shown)


antongolub:
@gr2m, could you take a look? We really need this fix asap. The current workaround is janky: we have to patch plugin inners to bypass npm registry rate limit.

travi:
> We really need this fix asap.

i recommend that you fork and publish a version under you own scope if you need this quickly. we are a very small team of volunteer maintainers and have other priorities than those that impact your timeline. this looks like a small change, but we need to consider the implications that this could have on a process that has been working well for lots of users for a long time.

it appears that this is only a problem because you are using semantic-release in a monorepo, which is an unsupported use case. we recognize that there are tools that enable support on top of our core, but officially it is unsupported. we do want to make support possible through tools like the one your are using, but that support is a lower priority than our core functionality. 

further, does this change even fully solve the problem, or does it just lower the chance of the problem? will this problem resurface if your monorepo grows to have even more packages?

antongolub:
We really appreciate and we are grateful to the semrel team for all the work that you do.

1. Correct. We can continue unsafe and fragile monkey-patching. We can also fork the plugin. Let's just imagine. The problem affects all `msr` users with monorepos of ~10 packages and more. We'd like to save them from the need to reconfigure all of these projects. Moreover, some custom plugins are tightly bound / inherited from npm plugin, and the chain of changes for them becomes even larger.
2. Technically yes. We trigger `verify-auth` as many times as the number of packages we're going to release. I guess multiple `npm whoami` calls run into the limit on the npmjs registry side.
3. semrel does not provide monorepos support. That's clear and this is your vision, which we do not even dispute. That's why another volunteer community has been trying to implement this functionality on its own for several years. 
* [atlassian/lerna-semantic-release](https://github.com/atlassian/lerna-semantic-release)
* [pmowrer/semantic-release-monorepo](https://github.com/pmowrer/semantic-release-monorepo)
* [dhoulb/multi-semantic-release](https://github.com/dhoulb/multi-semantic-release)
I note with pride that we succeed in this.

gr2m:
fyi I filed a support ticked with GitHub about npm's responses in case of a rate limit. The ticket got escalated but I didn't hear back yet. I asked about how to recognize a rate-limitted response so that we can throttle/retry requests.

antongolub:
Nice! We're now experimenting with a [custom npm plugin](https://github.com/semrel-extra/npm/blob/master/src/main/js/index.js). Approach with `verifyConditions` promise memorization seems to be working.

gr2m:
I received a response from support

> You'll receive a response with status code 429 when there are too many requests. This will generally happen on the search endpoint, but also applies to package /downloads URLs.

travi:
> You'll receive a response with status code 429 when there are too many requests.

would be nice to understand if there is a `Retry-After` header included, which is [common with 429 responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429). would it be worth following up to clarify that detail? i assume they did not provide a link to specific documentation around this?

> This will generally happen on the search endpoint, but also applies to package /downloads URLs.

it sounds like this is occurring with the `whoami` request, which doesn't seem like it is specifically mentioned (since i assume that would hit a different endpoint than search or package-downloads). do you get the impression that this is applied more generally than that response suggests?

moroine:
Hi, do we have any update on this? I'm facing the same issue randomly using multi-semantic-release

PR #437, opened 2021-11-24
refactor(esm): converted the package to esm

### LINKED ISSUE #436: ES Module
Part of our efforts to migrate `semantic-release` to native ES Modules: https://github.com/semantic-release/semantic-release/issues/2133

travi:
fixes #436 

BREAKING CHANGE: `@semantic-release/npm` is now a native ES Module. It has named exports for each plugin hook (`verifyConditions`, `prepare`,`publish`, `addChannel`)


travi:
tests are now passing in node v18 after the conversion to esm. 

https://github.com/semantic-release/npm/pull/538#issuecomment-1356057959 blocks v19 from passing until we decide how to proceed there. if we can figure out how to get the local npm dependency to be used, we can likely defer handling that case specifically, especially since npm v9 is not yet LTS anyway.

gr2m (approved):

PR #444, opened 2021-12-22
fix: drop npm dependency

antongolub:
closes #434
relates #270

alternative #445

https://github.com/semantic-release/npm/issues/434#issuecomment-999870524
> Suggestion: remove npm dependency. I [still believe](https://github.com/semantic-release/npm/pull/270) that the plugin always invokes global `npm`, so this dependency is completely useless.
> 
> Have a look:
> 
> https://github.com/semantic-release/npm/blob/708c29bf735d731f6855b2425ba9f12b618873dd/lib/publish.js#L23
> 
> `npm` is called via `execa`. It is just a wrapper for `child_process.exec` by default. cp uses `$PATH` to find util ref, and it knows absolutely nothing about `node_modules/.bin/npm`.
> 
> ```
> git clone ... && npm i
> npm -v
> 6.14.13
> 
> node -e "console.log(require('execa').sync('npm', ['-v']).stdout)"
> 6.14.13
> 
> node -e "console.log(require('./node_modules/npm/package.json').version)"
> 7.24.2
> ```
> 
> if we want execa to call the plugins's own npm version, we should pass [preferlocal option](https://github.com/sindresorhus/execa#preferlocal).
> 
> ```
> node -e "console.log(require('execa').sync('npm', ['-v'], {preferLocal: true}).stdout)"
> 7.24.2
> ```

Issue #434, opened 2021-11-22
npm audit - moderate vulnerabilities in deep json-schema dependency


perry-mitchell:
Getting a handful of vulnerability warnings with this package when running `npm audit` on the latest version:

```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Moderate      │ json-schema is vulnerable to Prototype Pollution             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ json-schema                                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=0.4.0                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ @semantic-release/npm [dev]                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ @semantic-release/npm > npm > node-gyp > request >           │
│               │ http-signature > jsprim > json-schema                        │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://github.com/advisories/GHSA-896r-f27r-55mw            │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Moderate      │ json-schema is vulnerable to Prototype Pollution             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ json-schema                                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=0.4.0                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ @semantic-release/npm [dev]                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ @semantic-release/npm > npm > @npmcli/run-script > node-gyp  │
│               │ > request > http-signature > jsprim > json-schema            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://github.com/advisories/GHSA-896r-f27r-55mw            │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Moderate      │ json-schema is vulnerable to Prototype Pollution             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ json-schema                                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=0.4.0                                                      │
[...]

PR #445, opened 2021-12-23
fix: add `preferLocal` option to allow execa to use local npm version

antongolub:
relates #434, #270 
alternative #444

From execa@2.0.0, `preferLocal` is set to `false` by default. The plugin uses `^5.0.0` now.

#177, 2 Jul 2019, execa 1.0.0 → 2.0.2

https://github.com/sindresorhus/execa/releases/tag/v2.0.0
sindresorhus/execa#314
sindresorhus/execa@eb22ff7

### DIFF
--- a/lib/verify-auth.js
+++ b/lib/verify-auth.js
@@ -21,7 +21,7 @@ module.exports = async (npmrc, pkg, context) => {
       const whoamiResult = execa('npm', ['whoami', '--userconfig', npmrc, '--registry', registry], {
         cwd,
         env,
-        preferLocal: true
+        preferLocal: true,
       });
       whoamiResult.stdout.pipe(stdout, {end: false});
       whoamiResult.stderr.pipe(stderr, {end: false});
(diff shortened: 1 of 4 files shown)


travi:
thank you for the investigation here. since this gets the behavior back to what was intended, i think this is the appropriate step for now over #444

travi (approved):

lib/verify-auth.js
```diff
@@ -18,7 +18,7 @@ module.exports = async (npmrc, pkg, context) => {
 
   if (normalizeUrl(registry) === normalizeUrl(DEFAULT_NPM_REGISTRY)) {
     try {
-      const whoamiResult = execa('npm', ['whoami', '--userconfig', npmrc, '--registry', registry], {cwd, env});
+      const whoamiResult = execa('npm', ['whoami', '--userconfig', npmrc, '--registry', registry], {cwd, env, preferLocal: true});
```
travi, line 21:
```suggestion
      const whoamiResult = execa('npm', ['whoami', '--userconfig', npmrc, '--registry', registry], {
        cwd,
        env,
        preferLocal: true
      });
```

PR #482, opened 2022-05-13
fix: do not update workspaces when applying new version

KillianHmyd:
In npm@8 the command `npm version` added a new command [workspaces-update](https://docs.npmjs.com/cli/v8/commands/npm-version#workspaces-update) which is `true` by default.
This option cause the following error when we install workspaces using non-released version (in this case the version `0.0.0-semantic-release`) :
> npm ERR! notarget No matching version found for package@0.0.0-semantic-release.

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -9,7 +9,15 @@ module.exports = async (npmrc, {tarballDir, pkgRoot}, {cwd, env, stdout, stderr,
 
   const versionResult = execa(
     'npm',
-    ['version', version, '--userconfig', npmrc, '--no-git-tag-version', '--allow-same-version'],
+    [
+      'version',
+      version,
+      '--userconfig',
+      npmrc,
+      '--no-git-tag-version',
+      '--allow-same-version',
+      '--no-workspaces-update',
+    ],
     {
       cwd: basePath,
       env,


travi:
can you provide more context around the situation where you are seeing this? i've never seen this error in projects that do not have workspaces defined. note that semantic-release core does not officially support monorepos. are you using something in addition to semantic-release to enable functionality within a monorepo?

travi:
> in my case we use yarn workspaces

are you managing multiple packages in the same repository through the use of yarn workspaces? assuming so, what else are you using to make this work with semantic-release? semantic-release does not support that type of repository structure by itself

KillianHmyd:
Yes we manage multiple packages and we use `nx` to run semantic-release in a mono-repo (cf their [example](https://github.com/arenglish/nx-semantic-release-demo)).
But, please note that here, the issue would be the same even if only one workspace were managed by semantic-release. Here the issue is occurs as long as some workspaces are declared in a repository, even if they are not managed via semantic-release.

KillianHmyd:
Hello @travi,
Will this PR be considered? Or do you plan to refuse?

KillianHmyd:
Hello @travi, sorry for the long delay from my side.
I didn't thought about it as I use `yarn` and not `npm`, I'll give it a try 😃

KillianHmyd:
@travi Setting it in the npmrc fix well the issue 👍🏾 Thanks for the suggestion

mario-jerkovic:
@KillianHmyd can you please share the content of your .npmrc, did you just wrote `no-workspaces-update` or ?

hongaar:
@mario-jerkovic I was able to publish a monorepo with *semantic-release* using this `.npmrc` committed to the root of the project:

```
workspaces = true
workspaces-update = false
include-workspace-root = true
```

Still bumping into the issue that the root package can't have `private: true` which is common in monorepos, the plugin will skip the _publish_ step in that case.

PR #503, opened 2022-07-25
Revert "chore(deps): lock file maintenance"

travi:
Reverts semantic-release/npm#488


travi:
was checking to see if this would resolve the consistent failures on node v17, but it did not. it is unclear why tests passed when this commit was merged, but consistently fail now

Issue #495, opened 2022-06-10
no-workspaces-update option


johngeorgewright:
A recent change to npm v8 runs an update, to any project that includes a `"workspaces"` property, when using the `npm version` command inside one of the configured "workspaces". I'm using `@semantic-release/npm` in a monorepo using the Yarn3 and this is now failing as my packages are linked together with a protocol that is not supported by NPM (and even if it were, I'd rather run this update with `yarn` and not `npm`).

It looks like this feature can be [turned off with a flag](https://docs.npmjs.com/cli/v8/commands/npm-version#workspaces-update) `--no-workspaces-update` when running `npm version`. Could this perhaps be an option for @semantic-release/npm?

janpe:
We're having this same issue as well. @johngeorgewright did you manage to find a workaround for the time being?

johngeorgewright:
@janpe, I ended up using a slightly more custom approach, with [@semantic-release/exec](https://github.com/semantic-release/exec). You can see [my config file here](https://github.com/johngeorgewright/runtyping/blob/master/release.config.js).

travi:
please also refer to [this conversation](https://github.com/semantic-release/npm/pull/482#issuecomment-1158706304) about setting this option in the `.npmrc` file for the project. 

keep in mind that semantic-release does not officially support use in the context of a monorepo at this point, so we are also avoiding adding complexity to the project for needs that are specific to monorepos. we do want to avoid details that block semantic-release from being enhanced by community addons for use in such contexts, but the use of an `.npmrc` file to set this option should be enough to avoid a semantic-release specific option for this.

PR #512, opened 2022-09-11
fix: Execute npm version with --workspaces=false

aarne:
Some tools like https://github.com/qiwi/multi-semantic-release wrap semantic release. Unfortunately with newer versions of npm people started getting EUNSUPPORTEDPROTOCOL errors during version locking as described here: 

 - https://github.com/qiwi/multi-semantic-release/issues/60
 - https://github.com/dhoulb/multi-semantic-release/issues/129

This change just adds `--workspaces=false` to npm command during versioning. This parameter has not side effects on normal repos but fixes the issue for monorepo setups

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -9,7 +9,7 @@ module.exports = async (npmrc, {tarballDir, pkgRoot}, {cwd, env, stdout, stderr,
 
   const versionResult = execa(
     'npm',
-    ['version', version, '--userconfig', npmrc, '--no-git-tag-version', '--allow-same-version'],
+    ['version', version, '--userconfig', npmrc, '--no-git-tag-version', '--allow-same-version', '--workspaces=false'],
     {
       cwd: basePath,
       env,


travi:
while we dont want to prevent such tools from extending semantic-release, keep in mind that we dont officially support monorepos. therefore, we avoid adding behavior here that is specifically for supporting monorepos. command-line flags for npm can also be added to your `.npmrc` file within your project and is our recommendation for situations like this one.

manast:
I created a .npmrc with this content:
```
workspaces=false
```

however when I try to make the release I get this error now:
```
npm ERR! Can not use --no-workspaces and --workspace at the same time
```
Any ideas?

aarne:
we run a simple command `npm version 1.0.0 --no-git-tag-version --allow-same-version` in individual packages.

1. usually it fails with EUNSUPPORTEDPROTOCOL
2. with --workspaces=false (this pr) it works as expected
3. with suggested modification to .npmrc we get an error `Can not use --no-workspaces and --workspace at the same time`

We can just create a new fork and publish `sr-npm-monorepo-safe` but that sounds a bit too much for such a simple change :) 

Most monorepo tools will allow you to switch out the implementation in release config

```
    "plugins": [
        "@semantic-release/commit-analyzer",
        "sr-npm-monorepo-safe",
    ]
```

manast:
> while we dont want to prevent such tools from extending semantic-release, keep in mind that we dont officially support monorepos.

@travi it could even be argued that since you do not support monorepos, adding workspace=false as default makes more sense than letting it be.

manast:
@aarne I made the change of this PR in a custom fork and the error I am getting now is this one?
```
npm ERR! Cannot read properties of null (reading 'matches')
```
Could not find any reference to this using google either. Locally I do not receive this error however, only when running as github actions.

aarne:
For me it's working with the mentioned modifications with @qiwi/multi-semantic-release. Are you sure this error is related to this change and this repo or its a general question? We can sync offline from here if its general to keep this PRs clear of unrelated discussions.

manast:
@arne I think this discussion is still relevant for this PR. I discovered that the last error is gone if I downgrade Node to version 14 instead of 16.

aarne:
@travi Please let us know if it makes sense to wait for this PR or we should be looking for another alternative?

travi:
other semantic-release users have been using workspaces successfully without changes to semantic-release, but with only `.npmrc`. i am not yet convinced that the situation you are describing is not in the same category. please see the following conversations:
* https://github.com/semantic-release/npm/pull/482#issuecomment-1158706304
* https://github.com/semantic-release/npm/issues/495

> it could even be argued that since you do not support monorepos, adding workspace=false as default makes more sense than letting it be.

the error that you mention above shows why this is not the case. if we include arbitrary flags in the commands executed by core, it increases the chances of conflicts with other flags needed by some projects

> 3\. with suggested modification to .npmrc we get an error `Can not use --no-workspaces and --workspace at the same time`

the conflicting flag is coming from something in your project, not semantic-release core. i suggest you find what is introducing the additional flag and determine if that flag is necessary.

again, i want to highlight that semantic-release itself _does not support monorepos_. a big part of this stance is because we do not have the bandwidth to dig into problems that are out of scope in addition to the other project priorities that are out in scope. we dont want to prevent other tools from extending semantic-release to add functionality like this, but the investigation and work needs to be done by the community around those tools. for us to even consider making a change to core that is specifically to enable monorepo usage, it needs to be proven that a change is needed that cannot be accomplished outside of core and that the change does not introduce problems for projects that are not using such an approach.

aarne:
@travi You are correct, adding `workspaces-update = false` avoids the configuration conflict and resolves the issue with version command. The additional references you sent helped, thank you.

I will update the still open bugs in other projects with instructions so others will find the solution also.

travi:
thank you for following up to let us know and for updating the other issues, @aarne. would you be interested in helping us make an update to our docs to help others find this information?

Issue #470, opened 2022-04-09
incompatible with pnpm monorepo


NullVoxPopuli:
I recently converted a monorepo from yarn to pnpm, and now publishing won't happen.

```
4:36:23 PM] [semantic-release] [@semantic-release/npm] › ℹ  Verify authentication for registry https://registry.npmjs.org
[4:36:23 PM] [semantic-release] [@semantic-release/npm] › ℹ  Reading npm config from /home/nullvoxpopuli/.npmrc
npm ERR! code ENOWORKSPACES
npm ERR! This command does not support workspaces.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/nullvoxpopuli/.npm/_logs/2022-04-09T20_36_24_000Z-debug-0.log
[4:36:24 PM] [semantic-release] › ✖  Failed step "verifyConditions" of plugin "@semantic-release/npm"
[4:36:24 PM] [semantic-release] › ℹ  Start step "verifyConditions" of plugin "@semantic-release/github"
```

browniebroke:
I have the same error with npm workspaces. The traceback points at this line: https://github.com/semantic-release/npm/blob/4b33e5bc8e051822aa79654e76bdd82c2f0452cd/lib/verify-auth.js#L21

Also, I can reproduce locally by logging in, moving into one of the workspace of the monorepo and running: `npm whoami`

browniebroke:
And it seems to be fixed in `npm@8.6` (assume it was part of https://github.com/npm/cli/pull/4479)

manast:
I also have got this error, is there any workaround for this, can we just force the use of npm@8.6 on the plugin?

hongaar:
Still have this problem with npm 8.19.2:

```
node@9a0111889c1a: /workspaces/mokr on main [$]
$ npm -v
8.19.2

node@9a0111889c1a: /workspaces/mokr on main [$]
$ npm whoami
npm ERR! code ENOWORKSPACES
npm ERR! This command does not support workspaces.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/node/.npm/_logs/2022-11-24T10_59_40_560Z-debug-0.log

node@9a0111889c1a: /workspaces/mokr on main [$]
$ cat .npmrc 
workspaces = true
workspaces-update = false
```

Opened https://github.com/semantic-release/npm/pull/529 to address this

jedwards1211:
[...]

PR #529, opened 2022-11-24
fix: add --no-workspaces to npm whoami

hongaar:
I'm using semantic-release with a monorepo which is not officially supported but discussed (e.g. https://github.com/semantic-release/semantic-release/issues/1688).

I've tried [various alternative](https://dev.to/antongolub/the-chronicles-of-semantic-release-and-monorepos-5cfc) but each of them gave me an uncomfortable feeling and after reading through https://github.com/semantic-release/npm/pull/482 I settled with the seemingly simple option of only using semantic-release together with this `.npmrc`: 

```
workspaces = true
workspaces-update = false
```

There's currently one issue blocking me from using this approach. Executing `npm whoami` with the above `.npmrc` yields an error, described in https://github.com/semantic-release/npm/issues/470.

This issue could be resolved by modifying `lib/verify-auth.js` slightly:

```diff
-      const whoamiResult = execa('npm', ['whoami', '--userconfig', npmrc, '--registry', registry], {
+      const whoamiResult = execa('npm', ['whoami', '--userconfig', npmrc, '--registry', registry, '--no-workspaces'], {
```

The [NPM docs](https://docs.npmjs.com/cli/v9/using-npm/config#workspaces) mention:

> Explicitly setting this to false will cause commands like install to ignore workspaces altogether

I have verified that this flag has no adverse effect, running it on monorepos and non-monorepos both with and without the flag yield the same output.

### DIFF
--- a/lib/verify-auth.js
+++ b/lib/verify-auth.js
@@ -18,7 +18,7 @@ module.exports = async (npmrc, pkg, context) => {
 
   if (normalizeUrl(registry) === normalizeUrl(DEFAULT_NPM_REGISTRY)) {
     try {
-      const whoamiResult = execa('npm', ['whoami', '--userconfig', npmrc, '--registry', registry], {
+      const whoamiResult = execa('npm', ['whoami', '--userconfig', npmrc, '--registry', registry, '--no-workspaces'], {
         cwd,
         env,
         preferLocal: true,


DerYeger:
This would be a welcome change, as I'm also blocked by the current behavior.

hongaar:
Since I use `yarn` for all my (mono)repos, I created [semantic-release-yarn](https://github.com/hongaar/semantic-release-yarn) which supports monorepos natively

jedwards1211:
@travi @gr2m this seems like a good idea, is there anything I could do to help get this over the line?

travi:
we are pretty strongly against adding code to core or official plugins that is specifically for projects that are monorepos. we are happy to better document workarounds like using npmrc, publishConfig in the package.json, or npm environment variables that can enable these sorts of things without changes to semantic-release. if someone knows a solution to this issue without adding a flag, please let us know. 

adding things like this add extra complexity to the project that we have to maintain and not break over time that are only there for unsupported use cases. we have a hard enough time keeping up with the use cases we do officially support. we do try to avoid doing things specifically to prevent use in monorepos, but we do not support them officially so priority is very low if changes are needed to the official projects to enable that to work beyond what the community provides to extend the official projects.

jedwards1211:
@travi I see.  As far as this issue goes, it seems like `npm` behavior is still wrong...waiting to hear confirmation from them that `whoami` shouldn't be able to work in workspaces.

FWIW, I've been using `semantic-release` for `pnpm` monorepos in my work for awhile now, without any special plugins.  The only fix I needed to get it working was https://github.com/semantic-release/npm/pull/531, and I think that fix is justified even for single package repo use cases.  (I didn't run into this issue because the `workspaces` npm config option isn't a requirement for working with monorepos.)  The only thing that was really difficult was configuring the `conventional-changelog` tools to distinguish commits by subpackage scope.  Their config format isn't as powerful as I'd like so [it wasn't pretty](https://github.com/jcoreio/toolchains/blob/main/packages/semantic-release/release.config.cjs#L41), but I at least got it working.

PR #534, opened 2022-12-07
feat: additional arguments `publishArgs` and `versionArgs` for npm commands

1aron:
- #512 
- #529

I built my own [monorepo ecosystem](https://github.com/1aron/aronrepo/tree/beta/packages/semantic-release-config) based on semantic release.

During the NPM release process, I encountered resistance, so I added two conservative features.

Since specifying a workspace in `.npmrc` such as `workspace=packages` will change the behavior of executing `npm run` in the root directory, we usually don't want to store `.npmrc`, and NPM will not automatically read the `.workspaces` settings in package.json when running `npm publish` and `npm version`, which is quite inconvenient.

Of course, we can execute `echo -e "\nworkspace=packages/**/*" >> .npmrc` on CI, but as mentioned above, it will affect the execution of subsequent instructions, which will make the command of the project not predictable.

Before submitting this PR I removed `@semantic-release/npm` package and solved my problem in `publishCmd`:
```json
{
     "publishCmd": "npm publish --workspace=packages"
}
```
But I found that I was missing a lot of the benefits that `@semantic-release/npm` brought me.

You can see the use cases I added in the README:
```json
{
     "publishArgs": ["--workspace", "packages"],
     "versionArgs": ["--workspace", "packages"]
}
```

These test cases were failing before I made any changes:

Screenshot 2022-12-08 at 12 07 57 AM

I even use it on the [Master CSS](https://github.com/master-co/css) open source project, which is a monorepo, and it works nicely with Semantic Release.

This is not a monorepo-specific feature, but this can be very useful in monorepo projects with multiple packages. Hope this feature passes quickly, thanks to the Semantic Release team for all you do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ Use either `NPM_TOKEN` for token authentication or `NPM_USERNAME`, `NPM_PASSWORD
 | `npmPublish` | Whether to publish the `npm` package to the registry. If `false` the `package.json` version will still be updated.  | `false` if the `package.json` [private](https://docs.npmjs.com/files/package.json#private) property is `true`, `true` otherwise. |
 | `pkgRoot`    | Directory path to publish.                                                                                          | `.`                                                                                                                              |
 | `tarballDir` | Directory path in which to write the package tarball. If `false` the tarball is not be kept on the file system. | `false`                                                                                                                          |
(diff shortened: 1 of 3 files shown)


1aron:
I've published my changes as `aron-semantic-release-config` and it works perfectly fine in practice:

Screenshot 2022-12-08 at 1 39 25 PM

https://github.com/1aron/aronrepo/actions/runs/3645164693/jobs/6155064392#step:6:275

PR #538, opened 2022-12-17
fix(registry-auth): scoped the legacy token in the npmrc to the registry

### LINKED ISSUE #530: Add support for npm v9
npm recently released v9 which includes some breaking changes in which the legacy authentication types have been consolidated into “legacy”, making the auth-type default to “web”. 

JFrog has an article describing the change https://jfrog.com/knowledge-base/artifactory-changes-to-the-login-behavior-in-npm-v9/.

I'm getting the error below using `npm@9.1.1`
```
[semantic-release] ��� ���  An error occurred while running semantic-release: Error: Command failed with exit code 1: npm version 1.2.35 --userconfig /tmp/7936e26bc5b88d9c14492113eacaa2fd/.npmrc --no-git-tag-version --allow-same-version
npm ERR! code ERR_INVALID_AUTH
npm ERR! Invalid auth configuration found: `_auth` must be renamed to `//registry.npmjs.org/:_auth` in user config
npm ERR! Please run `npm config fix` to repair your configuration.`
```

Can the `.npmrc` setup be modified to support npm v9?

travi:
fixes #530

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,8 @@ jobs:
         node-version:
           - 14.17
           - 16.0.0
-          - 17
+          - 18.0.0
+          - 19
         os:
           - ubuntu-latest
     runs-on: "${{ matrix.os }}"
--- a/lib/set-npmrc-auth.js
+++ b/lib/set-npmrc-auth.js
@@ -32,7 +32,9 @@ module.exports = async (
   if (NPM_USERNAME && NPM_PASSWORD && NPM_EMAIL) {
     await outputFile(
       npmrc,
-      `${currentConfig ? `${currentConfig}\n` : ''}_auth = \${LEGACY_TOKEN}\nemail = \${NPM_EMAIL}`
+      `${currentConfig ? `${currentConfig}\n` : ''}${nerfDart(
+        registry
+      )}:_auth = \${LEGACY_TOKEN}\nemail = \${NPM_EMAIL}`
     );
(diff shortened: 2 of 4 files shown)


travi:
it looks like this only [became scope-able to a registry in npm v8](https://docs.npmjs.com/cli/v8/using-npm/config?v=true#_auth), so this would not be a backward compatible change for all of the node versions currently supported by this plugin. 

since the official registry has not supported this legacy approach to auth [for some time](https://blog.npmjs.org/post/160809090595/basic-auth-to-be-limited-soon.html) and [registries other than the official registry require legacy auth to be scoped to a registry](https://docs.npmjs.com/cli/v8/using-npm/config?v=true#_auth), this would only impact use with projects targeting alternative registries that are not yet using npm v8.

quite honestly, since artifactory was the primary reason that legacy auth support was added to semantic-release in the first place, and [artifactory does now support token auth](https://www.jfrog.com/confluence/display/RTF/Npm+Registry#npmRegistry-Usingnpmlogin), i lean toward removing legacy auth support altogether rather than fixing it for npm v9.

gr2m:
> since [artifactory was the primary reason](https://github.com/semantic-release/npm/pull/10) that legacy auth support was added to semantic-release in the first place, and [artifactory does now support token auth](https://www.jfrog.com/confluence/display/RTF/Npm+Registry#npmRegistry-Usingnpmlogin), i lean toward removing legacy auth support altogether rather than fixing it for npm v9.

Agree 💯

PR #561, opened 2023-01-08
fix(revert): "fix(deps): update dependency npm to v9"

travi:
This reverts commit https://github.com/semantic-release/npm/commit/28265ec20bb230c4bb26f541508afb62272d5404.


travi:
thankfully this didnt release because our pipeline needs the node version updated for releases, but this further highlights that the local npm version is not being used as intended (this would cause test failures for legacy auth if it were). 

anyway, i dont think we would want to upgrade to non-lts npm versions even if everything were legitimately passing.

gr2m (approved):

PR #579, opened 2023-02-16
fix(deps): upgraded npm to v9

travi:
BREAKING CHANGE: the direct dependency on npm has been upgraded to v9. details of breaking changes can be found at https://github.com/npm/cli/releases/tag/v9.0.0


travi:
since npm v9 is now included in node v18, i think it is safe to get this upgraded. since we already have this beta going, this felt like the right time since it also makes it easy to call it another breaking change.

gr2m (approved):
🚀

PR #581, opened 2023-02-23
convert the plugin to esm along with other breaking changes

travi:


gr2m (approved):
🚀

travi:
i will be stepping away for a couple of weeks, so i plan to promote this to stable after time away

PR #582, opened 2023-02-24
ci(provenance): enabled provenance for publishing

travi:
from my experience so far, this should be all that is necessary. since we already have this beta in progress, we might as well try it here first.


travi:
looks like our integration tests complicate getting this added. converting to draft for now until i can get that sorted out

PR #614, opened 2023-05-08
docs(provenance): make the provenance details easier to consume

lirantal:
# What
Extending the npm provenance section to be more helpful and provide complete example.

# Why
CI/CD pipeline failed and required digging into the issue log to find out other permissions are also needed to release a GitHub version, etc.

# The change
Updates the npm provenance section with copy/paste working example of YAML code that helps developers not fall into missing permission problems when wanting to enable the feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,8 @@ to be enabled on the job:
       id-token: write # to enable use of OIDC for npm provenance
 ```
 
-It's worth noting that if you are using semantic-release to its fullest with a GitHub release, GitHub comments
-and other features then more permissions are required to be enabled on this job:
+It's worth noting that if you are using semantic-release to its fullest with a GitHub release, GitHub comments,
+and other features, then [more permissions are required](https://github.com/semantic-release/github#github-authentication) to be enabled on this job:
 
 ```yaml
     permissions:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +45,30 @@ Since semantic-release wraps the npm publish command, configuring provenance is
...
+For package provenance to be signed on the GitHub Actions CI the following permission is required
+to be enabled on the job:
+
+```yaml
+    permissions:
+      id-token: write # to enable use of OIDC for npm provenance
+```
+
+It's worth noting that if you are using semantic-release to its fullest with a GitHub release, GitHub comments
+and other features then more permissions are required to be enabled on this job:
```
travi, line 59: I think it could be worth linking to where these details are defined in the docs for our github plugin. what do you think?
```suggestion
It's worth noting that if you are using semantic-release to its fullest with a GitHub release, GitHub comments,
and other features, then [more permissions are required](https://github.com/semantic-release/github#github-authentication) to be enabled on this job:
```

travi:
also, looks like this fails our lint check. would you mind adjusting with a quick `npm run lint:fix`?

travi (changes requested):
also, looks like this fails our lint check. would you mind adjusting with a quick `npm run lint:fix`?

README.md
lirantal: Yep, looks great!

lirantal:
Should be fixed now.

travi (approved):
thanks again, @lirantal!

PR #634, opened 2023-06-14
fix: add --no-workspaces to npm version command in prepare

quinton22:
Allow the prepare command to work in workspaces

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -13,7 +13,7 @@ export default async function (
 
   const versionResult = execa(
     "npm",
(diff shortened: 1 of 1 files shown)


travi:
it should be possible to define this flag in the `.npmrc` file for your project, or possibly even in `publishConfig` in your `package.json`. since semantic-release does not officially support monorepos, we do not want to add details to our core that are specific to working in a monorepo context. 

more conversation around this can be found by [searching previous conversations](https://github.com/search?q=repo%3Asemantic-release%2Fnpm%20no-workspaces).

PR #745, opened 2024-01-31
chore: replace read-pkg with read-package-json-fast

mehulkar:


mehulkar:
@travi would you be able to help shepherd this in?

test/get-pkg.test.js
```diff
@@ -55,5 +55,5 @@ test("Throw error if package.json is malformed", async (t) => {
     errors: [error],
   } = await t.throwsAsync(getPkg({}, { cwd }));
 
-  t.is(error.name, "JSONError");
+  t.is(error.name, "JSONParseError");
```
mehulkar, line 58: This could technically be seen as a breaking change[^1]. The underlying libraries just use different error classes. 

- `read-pkg` uses [`parse-json`](https://github.com/sindresorhus/parse-json) which throws a [`JSONError`](https://github.com/sindresorhus/parse-json/blob/5ad65169677db8f164a32dcd08a19bc5ae8426e3/index.js#L6)
- `read-package-json-fast` uses [`json-parse-even-better-errors`](https://github.com/npm/json-parse-even-better-errors throws) [`JSONParseError`](https://github.com/npm/json-parse-even-better-errors/blob/67ea4ee317e231284b7e2304fc650b847c73ca52/lib/index.js#L74)

[^1]:  considering this is a `semantic-release` library, I'm guessing the change is breaking?

PR #746, opened 2024-02-01
feat: add pnpm support via `packageManager` option

### LINKED ISSUE #280: Allow to use pnpm as a publish utility
## The issue

I use [pnpm](https://pnpm.js.org/) for monorepo management and semantic-release for publishing new releases. 
[pnpm supports `workspace:` protocol](https://pnpm.js.org/en/workspaces#workspace-ranges-workspace) in package.json, so the packages in monorepo can be always linked to local versions, during `publish` phase this `woskpace:` protocol is removed. 

**The issue** is that semantic release uses `npm` cli tool for publishing which does not remove `workspace:` protocol and eventually the package is published with `workspace:` dependency which cannot be resolved

## Desired behavior

Allow to specify either the string template for publishing package or function in plugin options

## Workaround

In linux, I can use aliases and alias `npm` to `pnpm`

lars-berger:
Add new `packageManager` option (either `npm` or `pnpm`) which changes the package manager used to publish the package.

Closes #280


travi:
This is the npm plugin. It is strictly for use in projects that target publishing with npm. We have no intention of supporting other package managers in this plugin. A better alternative would be to create a pnpm plugin. Since we have no upcoming plans to create an official plugin for that, a community supported plugin is our recommendation.

lars-berger:
@travi That's unfortunate to hear, especially since it's the most upvoted issue for this project.

For anyone who happens to come across this, the plugin is published as `@glzr/semantic-release-npm`.

Issue #763, opened 2024-03-11
NPM Audit Signatures issue on 11.0.3


nicholasgriffintn:
Hey, just tried publishing our package which does signature checks before going out to NPM.

Errored with the following on those signature checks:

```
1 package has an invalid attestation:

@semantic-release/npm@11.0.3 (https://registry.npmjs.org/)

Someone might have tampered with this package since it was published on the registry!
```

Seems to be fine with 11.0.2. Also, this version is showing checked on NPM, so not entirely sure on the issue here, maybe something to check out?

travi:
A change was made to what is stored in sigstore when generating provenance before that version of our npm package was published. The latest npm cli is updated to handle the updates details, but it's not included with node yet. See https://github.com/npm/cli/issues/7279 for more details.

nicholasgriffintn:
Ahh yeah I got this error too on our package after publishing, forgot to update.

Thanks for the link.

PR #764, opened 2024-03-14
ci(corepack): pinned the expected dev version of npm and explicitly used it for audit signatures

travi:


gr2m (approved):

travi:
yep, some context can be found here: https://github.com/semantic-release/npm/issues/763#issuecomment-1989542327

basically need to update to the latest npm cli to get `audit signatures` working for provenance on recently published package versions. we chose to use corepack in this case, but other options are available, like simply updating to the latest as a step in the workflow file

Tbhesswebber:
> yep, some context can be found here: https://github.com/semantic-release/npm/issues/763#issuecomment-1989542327
> 
> basically need to update to the latest npm cli to get `audit signatures` working for provenance on recently published package versions. we chose to use corepack in this case, but other options are available, like simply updating to the latest as a step in the workflow file

Ah, thank you! I don't know why that didn't come up as I was searching.

PR #767, opened 2024-03-20
feat: Add 'workspaces' option for version upgrade

zhaotao1989:
Adds the 'workspaces' option as part of the version upgrade feature. This allows users to manage multiple packages within a single repository more efficiently. This enhancement improves the project's scalability and organization.

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -4,16 +4,22 @@ import { execa } from "execa";
 
 export default async function (
   npmrc,
-  { tarballDir, pkgRoot },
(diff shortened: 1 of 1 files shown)


travi:
this has been asked for and answered [several times in the past](https://github.com/search?q=repo%3Asemantic-release%2Fnpm+workspaces&type=issues). please search before re-requesting.

we are not interested in adding monorepo-specific complexity when monorepos are not officially supported. this can be configured using npm config approaches, like an npmrc, rather than needing a change here

Issue #784, opened 2024-04-28
Security Issue with out of date tar@6.2.0 found with SNYK


jebarpg:
Fixed in: @6.2.1 | Exploit maturity: MEDIUM

Overview
[tar](vscode-file://vscode-app/e:/Users/J-Dog/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) is a full-featured Tar for Node.js.

Affected versions of this package are vulnerable to Uncontrolled Resource Consumption ('Resource Exhaustion') due to the lack of folders count validation during the folder creation process. An attacker who generates a large number of sub-folders can consume memory on the system running the software and even crash the client within few seconds of running it using a path with too many sub-folders inside.

Remediation
Upgrade tar to version 6.2.1 or higher.

@semantic-release/npm@12.0.0
│ └─┬ npm@10.5.0
│   ├─┬ cacache@18.0.2
│   │ └── tar@6.2.0 deduped
│   ├─┬ libnpmdiff@6.0.7
│   │ └── tar@6.2.0 deduped
│   ├─┬ node-gyp@10.0.1
│   │ └── tar@6.2.0 deduped
│   ├─┬ pacote@17.0.6
│   │ └── tar@6.2.0 deduped
│   └── tar@6.2.0
├─┬ cacache@18.0.2 extraneous
│ └── tar@6.2.0 deduped
├─┬ libnpmdiff@6.0.7 extraneous
│ └── tar@6.2.0 deduped
├─┬ node-gyp@10.0.1 extraneous
│ └── tar@6.2.0 deduped
├─┬ pacote@17.0.6 extraneous
│ └── tar@6.2.0 deduped
└── tar@6.2.0 extraneous

Seems like npm needs to be updated to a higher version to resolve this issue.

travi:
first of all, for security related concerns, _please_ follow [our security policy](https://github.com/semantic-release/npm/security/policy) in the future.

> npm needs to be updated to a higher version

you can already make this update in your project with no change to this project since we depend on npm as a range, rather than as a specific version. this is also one of the reasons we recommend installing and running with npx rather than capturing semantic-release as a dependency in your project. in your case, your lockfile is resulting in pulling old version when there is already a fix available for this concern.

since this is already solvable without us releasing an update, I'm going to close this with the recommendation to update your lockfile or to switch to using npx instead.

jebarpg:
[...]

PR #785, opened 2024-04-28
fix(security): updated npm to 10.6.0 to fix security issue with tar

jebarpg:
Updated npm to 10.6.0 to fix security issue found with SNYK. All tests still passing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,19 +19,19 @@
     "Gregor Martynus (https://twitter.com/gr2m)"
   ],
   "dependencies": {
-    "@semantic-release/error": "^4.0.0",
-    "aggregate-error": "^5.0.0",
-    "execa": "^8.0.0",
-    "fs-extra": "^11.0.0",
-    "lodash-es": "^4.17.21",
-    "nerf-dart": "^1.0.0",
-    "normalize-url": "^8.0.0",
-    "npm": "^10.6.0",
-    "rc": "^1.2.8",
-    "read-pkg": "^9.0.0",
-    "registry-auth-token": "^5.0.0",
-    "semver": "^7.1.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@semantic-release/npm",
   "description": "semantic-release plugin to publish a npm package",
-  "version": "0.0.0-development",
+  "version": "1.0.0",
```
travi, line 4: we let semantic-release track the version with tags, so we intentionally do not maintain the version here. this is also not the correct version based on what has been released for this package

package.json
```diff
@@ -19,19 +19,19 @@
...
+    "fs-extra": "11.2.0",
+    "lodash-es": "4.17.21",
+    "nerf-dart": "1.0.0",
+    "normalize-url": "8.0.1",
+    "npm": "10.6.0",
+    "rc": "1.2.8",
+    "read-pkg": "9.0.1",
+    "registry-auth-token": "5.0.2",
+    "semver": "7.6.0",
+    "tempy": "3.1.0"
```
travi, line 34: we intentionally do not pin these versions in order to enable consumers to update to newer versions without needing to wait for a release of this package and also to enable npm to dedupe more effectively

travi (changes requested):
in the future, please limit a dependency update PR to only changing the indended package without making changes to other project conventions as part of the same change

travi:
as mentioned in https://github.com/semantic-release/npm/issues/784, the npm dependency can be updated without requring a change in this package. instead of releasing a change here, we recommend updating project lockfiles or switching to using npx to avoid using stale versions

PR #844, opened 2024-07-30
fix: respect scoped registries from publishConfig

### LINKED ISSUE #219: Use CLI --@scope:registry flag to override scoped registry defined in .npmrc
I'm using `semantic-release-npm` to publish a package on my Nexus registry. My package is scoped. 

I have the following config my `.npmrc` that defines a rule for installing my scoped package (get it from Nexus proxy repository): 

```BASH
@myscope:registry=https://my-nexus-proxy-npm-group-url
```

If I use npm cli to publish my package, I am able to override the `.npmrc` : 

- with the `--@scope:registry` arg (and like mentioned [here](https://github.com/npm/npm/issues/10117)):

```BASH
npm publish --@scope:registry=my-nexus-registry-URL
```
- or with adding this config to my `package.json`

```javascript
"publishConfig": {
    "@myscope:registry": "my-nexus-registry-URL"
  }
```

I am not able to achieve the same result with `semantic-release-npm`. The command that it uses to publish doesn't support the arg `--@scope:registry`, only the `--registry` arg is supported: 
```BASH
npm publish --userconfig /path/to/.npmrc --registry URL
```
Is there any way that I can configure `semantic-release` to use a scoped registry instead? Even when I try to configure the `publishConfig` attribute in my `package.json` like this 

```javascript
"publishConfig": {
    "@myscope:registry": "my-nexus-registry-URL"
  }
```
it's not working because the [`get-registry.js`](https://github.com/semantic-release/npm/blob/master/lib/get-registry.js) is taking into account only the `registry` attribute without supporting [the scoped registry config](https://docs.npmjs.com/misc/scope#associating-a-scope-with-a-registry).

Thanks.

Kenneth-Sills:
This stops a niche issue when using Gitlab registries, where setting the `.npmrc` to pull packages from the instance-wide registry while also setting the `package.json:publishConfig` to push to the project-local registry for a given `@scope` would pass the incorrect `--registry` flag to `npm publish` as well as associate the `NPM_TOKEN` with the wrong URL in the generated userconfig file.

This DOES NOT resolve #219. So `--registry` is still technically ignored for scoped packages. See
[npm/cli#7043](https://github.com/npm/cli/issues/7043) for more details on that.

I made sure not to affect the priority of other resolutions to avoid any breaking changes. An additional test has been added for this specific case.

### DIFF
--- a/lib/get-registry.js
+++ b/lib/get-registry.js
@@ -2,12 +2,15 @@ import path from "path";
 import rc from "rc";
 import getRegistryUrl from "registry-auth-token/registry-url.js";
 
-export default function ({ publishConfig: { registry } = {}, name }, { cwd, env }) {
+export default function ({ publishConfig = {}, name }, { cwd, env }) {
+  const scope = name.split("/")[0];
+  const publishRegistry = publishConfig[`${scope}:registry`] ?? publishConfig.registry;
+
   return (
-    registry ||
+    publishRegistry ||
     env.NPM_CONFIG_REGISTRY ||
     getRegistryUrl(
-      name.split("/")[0],
+      scope,
       rc(
         "npm",
         { registry: "https://registry.npmjs.org/" },
--- a/test/get-registry.test.js
+++ b/test/get-registry.test.js
@@ -39,6 +39,26 @@ test('Get the registry configured in "NPM_CONFIG_REGISTRY"', (t) => {
   );
 });
 
+test('Get the registry configured in "publishConfig" for scoped package', async (t) => {
+  const cwd = temporaryDirectory();
+  await fs.appendFile(path.resolve(cwd, ".npmrc"), "@scope:registry = https://custom3.registry.com");
+  await fs.appendFile(path.resolve(cwd, ".npmrc"), "registry = https://custom4.registry.com");
+
+  t.is(
+    getRegistry(
+      {
+        name: "@scope/package-name",
+        publishConfig: {
+          registry: "https://custom6.registry.com",
+          "@scope:registry": "https://custom5.registry.com",
+        },
+      },
+      { cwd, env: { NPM_CONFIG_REGISTRY: "https://custom1.registry.com/" } }
+    ),
+    "https://custom5.registry.com"
+  );
+});
+
 test('Get the registry configured in ".npmrc" for scoped package', async (t) => {
   const cwd = temporaryDirectory();
   await fs.appendFile(path.resolve(cwd, ".npmrc"), "@scope:registry = https://custom3.registry.com");


travi:
Could you please provide a public reproduction of the problem? What you describe is the default npm behavior, which we do or best to align with already by mostly deferring to npm to handle. I've used such configuration myself without complication, so I'm skeptical of this being a problem

Kenneth-Sills:
Thanks for your time, @travi!

I don't have a way to reproduce publicly right now, as this is happening in a private self-hosted instance of Gitlab. But I can go into a little more detail.

To define NPM's behavior, assuming a package called `@demo/package`...

# Scopeless

```jsonc
// package.json
{
    "publishConfig": {
        "registry": "https://pkg.fake.io"
    }
}
```

```sh
# .npmrc
registry=https://rc.fake.io
```

```sh
# Command
npm publish $(pwd) --registry "https://cli.fake.io"

# Attempts to publish to 
# https://cli.fake.io
```

# Scoped in the `.npmrc`

```jsonc
// package.json
{
    "publishConfig": {
        "registry": "https://pkg.fake.io"
    }
}
```

```sh
# .npmrc
@demo:registry=https://rc.fake.io
```

```sh
# Command
npm publish $(pwd) --registry "https://cli.fake.io"

# Attempts to publish to 
# https://rc.fake.io
```

# Scoped in Both `.npmrc` and `publishConfig`

```jsonc
// package.json
{
    "publishConfig": {
        "@demo:registry": "https://pkg.fake.io"
    }
}
```

```sh
# .npmrc
@demo:registry=https://rc.fake.io
```

```sh
# Command
npm publish $(pwd) --registry "https://cli.fake.io"

# Attempts to publish to 
# https://pkg.fake.io
```

----

This plugin currently *only* discovers the `register` field of `publishConfig`, but it will never recognize a `@scope:registry` field. Meanwhile, if you look at the changeset, you'll see it *does* locate scoped `registry` entries in the project's `.npmrc`.

So there's no issue in the first two cases above, but on the last case - "Scoped in Both `.npmrc` and `publishConfig`" - this plugin deviates from NPM's behavior. When it sets up the userconfig, it will call `getRegistry`, which will fail to find `@demo:registry` in the `package.json:publishConfig`. Instead, it will return the scoped registry from `.npmrc` and then set up the environment's `NPM_TOKEN` for *that* environment instead of the registry `npm publish` will actually use and authorization will fail when attempting to push.

This change just adds the ability to locate those scoped registries in `publishConfig` and have them override the entries in `.npmrc`.

Kenneth-Sills:
Another way to confirm this change is needed is just to reset the changes to `lib/get-registry.js` and the run the new test:

```txt
[test:unit       ]   get-registry › Get the registry configured in "publishConfig" for scoped package
[test:unit       ] 
[test:unit       ]   test/get-registry.test.js:47
[test:unit       ] 
[test:unit       ]    46:                 
[test:unit       ]    47:   t.is(         
[test:unit       ]    48:     getRegistry(
[test:unit       ] 
[test:unit       ]   Difference (- actual, + expected):
[test:unit       ] 
[test:unit       ]   - 'https://custom6.registry.com'
[test:unit       ]   + 'https://custom5.registry.com'
[test:unit       ] 
[test:unit       ]   › file://test/get-registry.test.js:47:5
```

It's unable to see the scoped registry under `publishConfig`.

Kenneth-Sills:
@travi Would you happen to have some time you could look at this?

travi:
i'm still unconvinced this is necessary. please help me understand why you would ever need to define a scoped registry specifically under `publishConfig`. `publishConfig` is already scoped to the context of the package being published, so it will only ever impact the current package. the scoping then already depends on the name of the package. 

why does `publishConfig.registry` not already solve this usecase without us needing to add special handing for a scoped registry?

travi:
> just generally a good practice to be compatible with NPM

this is a goal of ours, but clearly we have some gaps. i am hesitant to add more complexity to our implementation in order to handle rare cases, because the more paths we implement, the more opportunities there are for us to be out of sync with npm. this is a big part of why i'm asking the questions that i am in this thread. 

to be open, i would be far more open to accepting an update that replaced our implementation with the actual implementation that npm uses to resolve these details. since the npm cli was decomposed into smaller packages within the last few years, i'm curious if that logic is exposed in a way that we could just consume it directly. would you be willing to investigate this?

Kenneth-Sills:
> would you be willing to investigate this?

Yeah, of course! Unfortunately, this path is still pretty complex in NPM and I don't think it'd be a particularly helpful approach.

The gist of it is:

- Their base command class [loads in a `Config`](https://github.com/npm/cli/blob/780afc50e3a345feb1871a28e33fa48235bc3bd5/lib/npm.js#L67).
- The project configuration then [flattens itself](https://github.com/npm/cli/blob/latest/workspaces/config/lib/index.js#L198) with a [specialized flattening function](https://github.com/npm/cli/blob/latest/workspaces/config/lib/definitions/index.js#L3).
- This is used to generate a [manifest](https://github.com/npm/cli/blob/780afc50e3a345feb1871a28e33fa48235bc3bd5/lib/commands/publish.js#L192).
- Then that manifest is [resolved](https://www.npmjs.com/package/npm-package-arg).
- So that it can finally be passed to [`pickRegistry`](https://www.npmjs.com/package/npm-registry-fetch).

It would be nice if we could abstract this by calling `npm config` with `execa`, but that doesn't respect `publishConfig` overrides[^1].

With that, I still think the PR as-is is the easiest and cleanest way to support this feature.

[^1]: `npm config ls` will have a small section at the end disclosing differing `publishConfig` settings, but the `--json` interface won't.

muratgozel:
hi there, i spent hours checking my gitlab token, environment and other stuff that i thought related but looks like it was semantic-release. in gitlab, you can't pull + publish packages if they are under same main group. (scope in npm terms). the messages above explain the issue very clear and im hoping to see semantic-release be fully compatible with gitlab too. (because its a great tool) i will continue to use github until then.

Kenneth-Sills:
@muratgozel Thanks for the report! 

If you'd still like to use Gitlab, there's a workaround where you can run `npm config -L project @scope:registry url` before Semantic Release runs and after `npm install`. This will inject the value into `.npmrc`, which this plugin is able to read. Just make sure that the `package.json` does *not* have `publishConfig.registry` or this plugin will attempt to use it instead (but `publishConfig[@scope:registry]` is fine).

Kenneth-Sills:
@travi Any chance you've reconsidered accepting this PR given the feedback and research?

travi:
i appreciate the additional research and information. this is on my list to revist and consider more thoroughly, but i cannot promise a timeline at this point

PR #853, opened 2024-08-27
fix: check that npm token has write access on package [WIP]

### LINKED ISSUE #4: Check package permission on the registry
Currently we only check if the user is authenticated on the configured registry.
But if a user set a valid`NPM_TOKEN`, that doesn't have the permission to deploy the current package the release will fail during the publish phase. Similarly if the package name is already taken the release will fail.

To avoid that we should check in the `verifyConditions` if the package exists, and if it does make sure the user has `read-write` access to the package.
Unfortunately the [npm-registry-client](https://github.com/npm/npm-registry-client) doesn't implement the method. So it would have to be done with `npm access ls-collaborators <@scope/package>`.

In addition it would be great, in case of a `restricted` package to be able to check if the organization is private and paid. But I'm didn't find a way to do that.
During publish if the organization is not private and paid, the registry returns a `402` error.

jedwards1211:
fix #4

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,11 +4,13 @@ import { temporaryFile } from "tempy";
 import getPkg from "./lib/get-pkg.js";
 import verifyNpmConfig from "./lib/verify-config.js";
 import verifyNpmAuth from "./lib/verify-auth.js";
+import verifyNpmRelease from "./lib/verify-release.js";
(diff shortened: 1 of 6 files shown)


jedwards1211:
closing for now, this approach won't work

PR #1017, opened 2025-10-16
promote to stable

### LINKED ISSUE #958: ensure support for OIDC with the official registry
context https://github.com/orgs/community/discussions/161015

travi:
<a href='https://hearts.dev/projects/3/users/travi'></a>

## todo before merging

- [x] handle dry-run of publish for releases that happen from a sub-directory
- [x] refine messaging for auth failure errors

## outstanding issues after this effort
* ~~add-channel will fail because of lack of OIDC support~~ this appears to no longer be true (update: [maybe still not fully in the clear](https://github.com/semantic-release/npm/issues/958#issuecomment-3418607816), with limited cases hitting issues that are on the registry side)
* ci verification will not cover node v24. resolving this may warrant switching test frameworks, which is beyond the scope of adding trusted publishing capabilities

fixes #958

### DIFF
--- a/lib/definitions/errors.js
+++ b/lib/definitions/errors.js
@@ -37,20 +37,22 @@ Your configuration for the \`pkgRoot\` option is \`${pkgRoot}\`.`,
 export function ENONPMTOKEN({ registry }) {
   return {
     message: "No npm token specified.",
-    details: `When not publishing through [trusted publishing](https://docs.npmjs.com/trusted-publishers) an [npm token](${linkify(
+    details: `When not publishing through [trusted publishing](https://docs.npmjs.com/trusted-publishers), an [npm token](${linkify(
       "README.md#npm-registry-authentication"
     )}) must be created and set in the \`NPM_TOKEN\` environment variable on your CI environment.
 
-Please make sure to create an [npm token](https://docs.npmjs.com/getting-started/working_with_tokens#how-to-create-new-tokens) and to set it in the \`NPM_TOKEN\` environment variable on your CI environment. The token must allow to publish to the registry \`${registry}\`.`,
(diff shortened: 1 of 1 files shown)


travi:
<a href='https://hearts.dev/projects/3/users/travi'></a>

@semantic-release/maintainers based on early feedback from https://github.com/semantic-release/npm/issues/958 and my tests in https://github.com/travi-test/npm-oidc-test/ beyond and our automated tests running in our pipeline, i'm ready to say this is ready for final review before promoting to stable. 

the outstanding issues in the initial PR description have been updated to remain accurate and are beyond the scope of what i think should hold this PR back from being merged

travi:
<a href='https://hearts.dev/projects/3/users/travi'></a>

> Is this the kinda pattern we get to follow if we want to support OIDC in other CI environments?? 🤔

unfortunately, yes. with the current state of things. any new ci providers that are added to the supported list on the npm registry side would require additional implementation in this plugin. i think this work sets us up for that work to be pretty minimal, but it would require work. 

the same would be true if alternative registries followed the path outlined in the "For other Registries" section of https://github.com/npm/cli/pull/8336. however, this work prepares us less for that scenario, since this currently puts a hard limit on the official registry. we'd need to check the registry for the presence of [this endpoint](https://api-docs.npmjs.com/#tag/registry.npmjs.org/operation/exchangeOidcToken) if that pattern were followed. if a different pattern were followed, we'd need an even more custom implementation per registry.

i don't like this plugin to have any knowledge of the ci service that it is running in, but in the case of both "trusted publishing" and provenance, the ci context is an important detail. ideally, the npm cli could give us a way to check if it is supported without needing to know the level of detail about the context that was required for this change, but that is the current reality. 

this was my reasoning when opening https://github.com/npm/cli/issues/8525. i think there are valid pieces of that request to follow up on after getting this version released. i wanted to be early in supporting trusted publishing, but that gap delayed getting this implementation together. now, there are tight deadlines that have already partially passed. this implementation works in the context of the current reality enabled by the current cli and registry. i hope we can influence that path to improve in the future after we release this version. it could simplify our current implementation, but also enable us to support future options as they become available.

PR #1122, opened 2026-03-31
feat: Generic & CircleCI Trusted Publishing

### LINKED ISSUE #1121: Support OIDC trusted publishing on CircleCI
<a href='https://hearts.dev/projects/3/users/dopry'></a>

CircleCI is a supported provider.  It looks like https://github.com/semantic-release/npm/blob/62e80ac3e33b8eac2584e52f5672886df2491b3d/lib/trusted-publishing/token-exchange.js#L60 could be updated with 

```
  if (CIRCLE_CI_PROVIDER_NAME === ciProviderName) {
    return exchangeCircleToken(pkg.name, logger); // identical to exchangeGitlabPipelinesToken except the log message.
  }
```
 as long as there is something like 
```
## 1. Fetch the OIDC token with the correct audience for npm
            export NPM_ID_TOKEN=$(circleci run oidc get --claims '{"aud": "npm:registry.npmjs.org"}')
```
in the circleci config. 

or possibly even better... Just check for NPM_ID_TOKEN and use it if present  the way `npm` does, only falling back to platform specific code on platforms when it is not present. 

```
export default function exchangeToken(pkg, { logger }) {
  const npmIdToken = process.env.NPM_ID_TOKEN;
   
  // if NPM_ID_TOKEN provided, give it precendence
  if (npmIdToken) { 
     logger.log("Trusted Publishing with NPM_ID_TOKEN env var");
     return exchangeToken(npmIdToken, pkg.name, logger);
  }

  // CI Provider Specific Handling
  const { name: ciProviderName } = envCi();

  if (GITHUB_ACTIONS_PROVIDER_NAME === ciProviderName) {
    return exchangeGithubActionsToken(pkg.name, logger);
  }
  return undefined;
}
```

dopry:
<a href='https://hearts.dev/projects/3/users/dopry'></a>

- add generic support for trusted publishing via NPM_ID_TOKEN to decouple semantic release from CI platforms.
- document trusted publishing with CircleCI

fixes #1121

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,28 @@ id_tokens:
 
 See the [npm documentation for more details about configuring pipeline details](https://docs.npmjs.com/trusted-publishers#gitlab-cicd-configuration)
 
+#### Trusted publishing for CircleCI
+
+To use trusted publishing in [CircleCI](https://docs.npmjs.com/trusted-publishers#circleci-configuration) you need to set the environment variable NPM_ID_TOKEN to an NPM Identity Token
+
+```yaml
+jobs:
+  semantic-release:
+    steps:
+      # ...
+      - run:
+          name: Semantic Release with Trusted Publishing via OIDC
+          command: |
+            # Fetch the OIDC token with the correct audience for npm
+            export NPM_ID_TOKEN=$(circleci run oidc get --claims '{"aud": "npm:registry.npmjs.org"}')
+            # Run semantic release
+            npx semantic-release
+```
+
+#### Generic Trusted Publishing
+
+The generic mechanism for telling semantic release and NPM to use trusted publishing is to set the NPM_ID_TOKEN. This generic approach is meant for use with CI platforms that have implemented trusted publishing support with NPM, but don't have specific support in semantic-release. If token exchange fails it will fall back to using Token authentication.
+
 #### Unsupported CI providers
 
 Token authentication is **required** and can be set via [environment variables](#environment-variables).
@@ -97,9 +119,10 @@ See the documentation for your registry for details on how to create a token for
 
 ### Environment variables
 
-| Variable    | Description                                                                                                                   |
-| ----------- | ----------------------------------------------------------------------------------------------------------------------------- |
-| `NPM_TOKEN` | Npm token created via [npm token create](https://docs.npmjs.com/getting-started/working_with_tokens#how-to-create-new-tokens) |
+| Variable       | Description                                                                                                                                                                                                             |
+| -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `NPM_TOKEN`    | Npm token created via [npm token create](https://docs.npmjs.com/getting-started/working_with_tokens#how-to-create-new-tokens)                                                                                           |
+| `NPM_ID_TOKEN` | OIDC identity token for [trusted publishing](https://docs.npmjs.com/trusted-publishers). Must be configured in your CI job (see [GitLab](#trusted-publishing-for-gitlab-pipelines), [CircleCI](#trusted-publishing-for-circleci)). Takes priority over CI-specific token retrieval when set. |
 
 ### Options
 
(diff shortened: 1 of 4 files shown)


z4ce:
@travi any way you could consider merging this so those of us using semantic release on circle can use trusted publishing?

README.md
```diff
@@ -97,9 +119,10 @@ See the documentation for your registry for details on how to create a token for
...
-| `NPM_TOKEN` | Npm token created via [npm token create](https://docs.npmjs.com/getting-started/working_with_tokens#how-to-create-new-tokens) |
+| Variable       | Description                                                                                                                                                                                                             |
+| -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `NPM_TOKEN`    | Npm token created via [npm token create](https://docs.npmjs.com/getting-started/working_with_tokens#how-to-create-new-tokens)                                                                                           |
+| `NPM_ID_TOKEN` | OIDC identity token for [trusted publishing](https://docs.npmjs.com/trusted-publishers). Must be configured in your CI job (see [GitLab](#trusted-publishing-for-gitlab-pipelines), [CircleCI](#trusted-publishing-for-circleci)). Takes priority over CI-specific token retrieval when set. |
```
travi, line 125: > Must be configured in your CI job

i think we need some adjustment here to clarify that this is only needed for specific ci providers. the current wording doesnt make it clear that this is _not_ required in the context of github actions, for example

travi:
one of my hesitations with this approach is the generic handling. the npm docs [still only list three supported providers](https://docs.npmjs.com/trusted-publishers#supported-cicd-providers). suggesting that a generic approach can work beyond that list goes beyond what i think the npm team has defined. i'm ok with adding circle to our supported list, but i think i would still want to handle it specifically rather than generically.

i think the path to more generic support depends on what the npm team ends up doing with https://github.com/npm/cli/issues/8525 because i want to avoid us being in the game of keeping up with each additional ci provider enabled by the npm team

z4ce:
@dopry would you be able to modify the PR to address @travi 's concerns and scope it to just circle until the generic question is answered upstream?

README.md
dopry: @travi  what would you suggest?

dopry:
> one of my hesitations with this approach is the generic handling. the npm docs [still only list three supported providers](https://docs.npmjs.com/trusted-publishers#supported-cicd-providers). suggesting that a generic approach can work beyond that list goes beyond what i think the npm team has defined. i'm ok with adding circle to our supported list, but i think i would still want to handle it specifically rather than generically.
> 
> i think the path to more generic support depends on what the npm team ends up doing with [npm/cli#8525](https://github.com/npm/cli/issues/8525) because i want to avoid us being in the game of keeping up with each additional ci provider enabled by the npm team

I specifically did the 'generic' so the semantic-release team wouldn't need to chase every added CI provider and because that is the way npm works in practice, even if it's not documented that way.  Frankly, its saddening that NPM is picking specific vendors to support over creating open standards in this case.

README.md
chammond-tz: What if you just added "This is not required when using trusted publishing in Github Actions." after the parenthetical and before the "takes priority..."?

chammond-tz: @dopry and @travi, any thoughts on moving this forward?

travi: i'm good with @chammond-tz's suggestion or similar for this one. however, i'm still looking for the overall change to be focused specifically on circle ci and not generic. as mentioned in another comment, we are looking for the npm team to lead what the approach should be for more generic handling

travi:
> I specifically did the 'generic' so the semantic-release team wouldn't need to chase every added CI provider and because that is the way npm works in practice

i understand the goal of simplifying here, but the OIDC handshake is limited to approved trusted providers, so it isnt fully generic. until npm would release support for generic providers, which i dont think is likely or even technically possible, i dont think we should advertise generic support here. as mentioned in https://github.com/semantic-release/npm/pull/1122#issuecomment-4463519429, we need something to happen on the npm side of the equation in order for us to be able to simplify here

dopry:
No worries. I'm not interested in making an implementation that doesn't reflect the way npm itself works under the hood. Feel free to roll your own.

travi:
> the npm docs [still only list three supported providers](https://docs.npmjs.com/trusted-publishers#supported-cicd-providers). suggesting that a generic approach can work beyond that list goes beyond what i think the npm team has defined.

related: https://github.com/orgs/community/discussions/186343

